PR #30, opened 2021-01-11
Set primary color and fontsize in theme.js

alvyjudy:
primary color was set to the orange and base font size was set to 16px in theme.js

### DIFF
--- a/client/src/themes/theme.js
+++ b/client/src/themes/theme.js
@@ -3,12 +3,12 @@ import { createMuiTheme } from "@material-ui/core";
 export const theme = createMuiTheme({
   typography: {
     fontFamily: '"Roboto"',
-    fontSize: 12,
+    fontSize: 16,
     h1: {
       // could customize the h1 variant as well
     }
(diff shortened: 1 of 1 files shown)


client/src/themes/theme.js
```diff
@@ -3,12 +3,12 @@ import { createMuiTheme } from "@material-ui/core";
 export const theme = createMuiTheme({
   typography: {
     fontFamily: '"Roboto"',
-    fontSize: 12,
+    fontSize: 16,
```
sina-jamshidi: I'll approve for now but isn't 16 kind of big? lol

sina-jamshidi (approved):
you're good to merge!

client/src/themes/theme.js
alvyjudy: on second thought, it might actually be too big... I'll revert it to 12 😅

alvyjudy:
Ticket number #29  (Forgot to link)

PR #32, opened 2021-01-12
Feature/models

### LINKED ISSUE #4: BE: Meeting Model
Create Model for Event Type

Event Type
- user_id
- duration (15, 30, 45) -> these are in minutes

### LINKED ISSUE #5: BE: Appointment Model
Create Appointment Model

Appointment

- event_type id
- name 
- email 
- time (datetime) 
- timezone

WhoaNellie:
added Appointment and EventType models as laid out in the sprint.

I have some questions about what types ids and timezones should be, as well as if we should add an 'attendees' field or something on Appointment to associate with user_id


.gitignore
```diff
@@ -21,3 +21,4 @@ yarn-error.log*
 
 #TODO add .env to your gitignore
 
+/server/db/password.js
```
dproc96: I think it would be better to keep the password and similar secrets in a .env file that we ignore

server/db/connection.js
```diff
@@ -2,9 +2,11 @@ require("dotenv").config();
 const mongoose = require("mongoose");
 const db = require("./models");
 
+const { password } = require("./password");
```
dproc96: This will also need to be updated to process.env when you move the password to the .env

server/db/connection.js
```diff
@@ -18,6 +20,25 @@ try {
             //         db.Test.find().then((data) => console.log(data));
             //     }
             // );
+
+            // testing appointment model
```
dproc96: I think we can remove these commented out tests, unless you want to keep them for specific testing in the future and then remove them

server/db/models/Appointment.js
```diff
@@ -0,0 +1,34 @@
+const mongoose = require("mongoose");
+
+const Schema = mongoose.Schema;
+
+const schema = new Schema({
+    meeting_id: {
+        // TODO: generate unique meeting id
+        // unique: true,
+        type: String,
```
dproc96: This should probably be a Schema.Types.ObjectId so that we can link it directly to the meeting type. Also Mongoose will automatically generate a unique id for us

server/db/models/EventType.js
```diff
@@ -0,0 +1,22 @@
+const mongoose = require("mongoose");
+
+const Schema = mongoose.Schema;
+
+const schema = new Schema({
+    user_id: {
+        // Link to appointments using user_id?
+        type: Number,
```
dproc96: This should also probably be an ObjectId but maybe we can leave that until we know what the user setup will be

dproc96 (changes requested):
Looks good overall, just a couple of changes I noticed

server/db/models/Appointment.js
WhoaNellie: Ah, I thought meeting ids would be readable but unique to the user, like the calendly link with an id of something like '/team-sprint' you pointed out

dproc96: That's an interesting point, let's discuss in the standup

server/db/models/Appointment.js
```diff
@@ -0,0 +1,32 @@
+const mongoose = require("mongoose");
+
+const Schema = mongoose.Schema;
+
+const schema = new Schema({
+    meeting_id: {
+        type: mongoose.ObjectId,
```
dproc96:
```suggestion
        type: mongoose.Schema.Types.ObjectId,
```
This is preferred when defining a Schema

server/db/models/EventType.js
```diff
@@ -0,0 +1,21 @@
+const mongoose = require("mongoose");
+
+const Schema = mongoose.Schema;
+
+const schema = new Schema({
+    user_id: {
+        type: mongoose.ObjectId,
```
dproc96:
```suggestion
        type: mongoose.Schema.Types.ObjectId,
```

dproc96 (approved):
Be sure to make those two minor updates, then this looks ready to go to me!

PR #33, opened 2021-01-12
Front end part of onboarding page 

alvyjudy:
Ticket number #12 

# Summary

- Inside the `pages/OnBoarding`, the `index.js` is the entry point to the page. It imports the three components (url&timezone setting, google calendar confirmation, availability) required for the three onboarding steps. 
- Inside `pages/Onboarding/smallComponents/` there are three components to handle smaller details of the page (progress bar, date selection)
- On each page flip, the browser sends a request to the backend (for now it just logs to the console)
- I used refs hooks to gather data from deeply nested components and had the top level component send request, which I'm not sure if is the best option. An alternative would be to set up a context and dispatch action from below. Still debating 😕
- the color should be correct when the theme.js PR gets merged.

# What it looks like right now

https://user-images.githubusercontent.com/53921639/104300888-06731e00-5495-11eb-8111-e83bb3dd7180.mov

There is still apparently quite some work to do (fixing the grids that blew up, and just more work on the look of the components and data flow). I'm wondering if I can get some feedback/help on it first? Feeling a bit lost at the moment 😅


dproc96:
It might be a good idea to keep all of the components in a separate components folder stored at /src/components. This will make it easier to reuse components and find them later on in development. It may also be good to follow the convention of Onboarding.js rather than Onboarding/index.js

client/src/pages/OnBoarding/index.js
```diff
@@ -0,0 +1,88 @@
...
+import ConnectGoogleCalendar from "./ConnectGoogleCalendar";
+import SetAvailability from "./SetAvailability";
+
+const OnBoarding = (props) => {
+  //styles
+  const classes = useStyles();
+
+  //data related variables
+  const [pageNum, setPageNum] = useState(0);
+  const url = useRef();
```
dproc96: It would be easier to keep these as state objects and make handler functions that get passed down to the children, rather than worrying about keeping high order refs consistent with lower level state

client/src/pages/OnBoarding/index.js
```diff
@@ -0,0 +1,88 @@
...
+  const timezone = useRef();
+  const days = useRef();
+  const startHour = useRef();
+  const finishHour = useRef();
+
+  const FlipToNextPage = () => {
+    //send data to backend here
+    switch (pageNum) {
```
dproc96: We could probably use an array of handler functions or something similar in place of a switch statement since we're already using integers for the cases. Not a big problem but may be cleaner.

client/src/pages/OnBoarding/index.js
```diff
@@ -0,0 +1,88 @@
...
+        break;
+      default:
+        break;
+    }
+  };
+
+  return (
+    <Paper elevation={3} className={classes.paper}>
+      <div className={classes.mainContent}>
+        {pageNum === 0 && <SetTimezoneUrl url={url} timezone={timezone} />}
```
dproc96: Might be better for user experience if we use routes, like "/onboarding/1", in case the user presses back expecting it to lead to the previous page

client/src/pages/OnBoarding/smallComponents/AvailableHoursButton.js
```diff
@@ -0,0 +1,39 @@
+import React, { useState, useEffect } from "react";
+import { TextField, MenuItem } from "@material-ui/core";
+
+const timeSlots = [];
+for (let i = 0; i <= 24; i++) {
```
dproc96:
```suggestion
for (let i = 0; i < 24; i++) {
```
Right now this is causing the times to go past midnight, note that we should also append 24:00 at the end of the array when finished

client/src/pages/OnBoarding/smallComponents/AvailableHoursButton.js
```diff
@@ -0,0 +1,39 @@
...
+  for (let j = 0; j <= 60; j = j + 15) {
```
dproc96:
```suggestion
  for (let j = 0; j < 60; j = j + 15) {
```
This is making times show up as 2:60 for example

client/src/pages/OnBoarding/smallComponents/AvailableHoursButton.js
```diff
@@ -0,0 +1,39 @@
...
+    const hour = i < 10 ? "0" + i.toString() : i.toString();
+    const min = j < 10 ? "0" + j.toString() : j.toString();
+    timeSlots.push(hour + min);
```
dproc96, line 10:
```suggestion
    timeSlots.push(`${hour} : ${min}`);
```
This would save you some time further down

dproc96 (changes requested):
I like the look of the UI but there are a couple of things I noticed. Additionally some of the styling isn't displaying cleanly (like the bottom example), using something like a Grid or a Flexbox would be a good solution for this example at least. 

Screen Shot 2021-01-12 at 8 47 41 AM

For an issue like this, adding some margin between the items would be a good idea

Screen Shot 2021-01-12 at 8 48 50 AM

alvyjudy:
I have made more edits. Specifically:
- Each step has its own URL now (back button now works)
- Fix up more styling issues

Next step I will reduce the use of Refs and pass handler function directly into children per Dan's suggestion. 

https://user-images.githubusercontent.com/53921639/104385352-188aa600-5501-11eb-84d6-47676c1e6cc5.mov

client/src/pages/OnBoarding/OnBoarding.js
```diff
@@ -0,0 +1,94 @@
...
+const OnBoarding = (props) => {
+  //styles
+  const classes = useStyles();
+
+  //data related variables
+  const url = useRef();
+  const timezone = useRef();
+  const days = useRef();
+  const startHour = useRef();
+  const finishHour = useRef();
```
sina-jamshidi: replace with `useState`. `useRef` ends up being redundant since you also have local state. Pass the states and the setters to the appropriate components. Also keep in mind that React's [official recommendations for forms is to have them be controlled](https://reactjs.org/docs/uncontrolled-components.html#:~:text=In%20a%20controlled%20component%2C%20form,form%20values%20from%20the%20DOM), so you do want to keep using state for the various form elements.

client/src/pages/OnBoarding/OnBoarding.js
```diff
@@ -0,0 +1,94 @@
...
+  const match = useRouteMatch({ path: "/onboarding/:page" });
+  const page = match ? match.params.page : 0;
+
+  const FlipToNextPage = () => {
+    switch (page) {
+      case "1":
+        history.push("/onboarding/2");
+        break;
+      case "2":
+        history.push("/onboarding/3");
```
sina-jamshidi: Since we're using `react-router-dom`, let's be consistent and use their components. You're looking for [Link](https://reactrouter.com/web/api/Link), which can be wrapped around a button.
Another less correct option is `Redirect`.

client/src/pages/OnBoarding/OnBoarding.js
```diff
@@ -0,0 +1,94 @@
...
+        break;
+      default:
+        break;
+    }
+  };
+
```
sina-jamshidi, line 34: Try making a handler function here, or pass state setters to your components

client/src/pages/OnBoarding/smallComponents/ProgressBar.js
```diff
@@ -0,0 +1,41 @@
...
+  const classes = useStyles();
+
+  const start = props.start || 0;
+  const end = props.end;
+  const [activeStep, setActiveStep] = useState(start);
+
+  useEffect(() => {
+    setTimeout(() => {
+      setActiveStep(start + 1);
+    }, 100);
```
sina-jamshidi: why is this necessary?

sina-jamshidi (changes requested):

client/src/pages/OnBoarding/smallComponents/ProgressBar.js
alvyjudy: It has something to do with how I mounted the component (which wasn't really smart). But I refactored it so the setTimeout won't be necessary now:)

client/src/pages/OnBoarding/OnBoarding.js
alvyjudy: I realize it would be much easier to just have those three components manage their own data handling and thus no need for handler function here :P

alvyjudy:
I have made more edits per Sina's suggestion above. Each components now contain their own local state without Refs and the page switching is done with react-router's Link component instead of invoking the history API. The progress bar is also mounted persistently so I don't have to call a setTimeout for the animation to show.

alvyjudy:
I submitted more changes to the styles to bring it closer to the mock up. Here is what it looks like right now (thanks Taras for the style inspiration in your PR!)
image
image
image

client/src/pages/OnBoarding/smallComponents/AvailableDays.js
```diff
@@ -0,0 +1,80 @@
...
+};
+
+const useStyles = makeStyles((theme) => ({
+  daysGrid: {
+    margin: "0",
+  },
+  eachDay: {
+    margin: "0",
+  },
+  label: {
```
dproc96, line 76: You should add some padding and/or include a different set width so that the text has room to breath but the boxes are consistent

dproc96 (commented):
In addition to my comment below, did you have a chance to check out the TimePicker for Material UI? Might be an easier solution than the component you made

client/src/pages/OnBoarding/OnBoarding.js
```diff
@@ -0,0 +1,79 @@
...
+  Divider,
+} from "@material-ui/core";
+import { Route, Switch, Redirect, useRouteMatch } from "react-router-dom";
+
+import SetTimezoneUrl from "./SetTimezoneUrl";
+import ConnectGoogleCalendar from "./ConnectGoogleCalendar";
+import SetAvailability from "./SetAvailability";
+import ProgressBar from "./smallComponents/ProgressBar";
+
+const OnBoarding = (props) => {
```
dproc96: By removing the state from the parent component, you're probably making your life a lot harder when you need to start making requests to the backend. It would be better to store everything up here and then pass down handlers to update.

client/src/pages/OnBoarding/smallComponents/AvailableDays.js
alvyjudy: Yes I did check it out but it looks quite different from what I see on calendly. Time picker seems to only allow the option to display a clock pop up but in calendly they use a list of available time options. 

I rolled out my own components to better match what we have on the mock up but either approach is okay with me. Should I switch to time picker?

client/src/pages/OnBoarding/OnBoarding.js
alvyjudy: I was actually debating whether to keep the state from parent component. I thought that since each request sent to the backend would be slightly different, wouldn't it make more sense to have each child component handle their own state and request?

client/src/pages/OnBoarding/smallComponents/AvailableDays.js
dproc96: I think you can just stick with what you have in that case. One note though is to adjust the spacing for it because the -- looks a bit off center

dproc96:
This looks good but there are still some spacing issues on the Availability page, can you take a shot at those? Then this should be good to merge

alvyjudy:
I made more changes and fixed the styling in the set availability page, here is a snapshot: 
image

dproc96 (approved):
This looks good to go!

PR #35, opened 2021-01-13
Signup login pages

kozaktar:
**Login and Signup pages features:**
-setup react router routs for Login and Signup pages
-"/" now redirects to Signup page
-Login and Signup pages styled reasonably close to the mocks

**Missing features:**
-Sign up page "Prefer to signup with a password" click here link doesnt work as create profile page isn't finished yet.
-Login with google and Signup with google buttons don't work yet as backend login and signup is not done yet [this will be integrated once I've finished that functionality]

**Folder structure:**
-added assets folder under src/

**Theme:
login1
login2
signup1
signup2

**
-theme file modified to include typography.h4, typography.h5, typography.button, palette.button styles


client/src/pages/Login.js
```diff
@@ -0,0 +1,166 @@
...
+    event.preventDefault();
+    //cehck if user has entered an email
+    if (email) {
+      showWelcome(true);
+    }
+  };
+
+  const handleChange = event => {
+    setEmail(event.target.value);
+    console.log(email);
```
dproc96: You should generally remove console.log statements before merging (some exceptions on backend obviously)

client/src/pages/Login.js
```diff
@@ -0,0 +1,166 @@
...
+                style={{ textAlign: "center" }}
+                inputProps={{ min: 0, style: { textAlign: "center" } }}
+                onChange={handleChange}
+              />
+            </div>
+          )}
+          <Button
+            size="large"
+            className={classes.button}
+            onClick={handleClick}
```
dproc96: Instead of having an onClick, it would be better for accessibility to make this type="submit" and it will use the onSubmit for the form

client/src/pages/Signup.js
```diff
@@ -0,0 +1,175 @@
...
+                style={{ textAlign: "center" }}
+                inputProps={{ min: 0, style: { textAlign: "center" } }}
+                onChange={handleChange}
+              />
+            </div>
+          )}
+          <Button
+            size="large"
+            className={classes.button}
+            onClick={handleClick}
```
dproc96: Same comment as the LogIn page for this one, could be a submit button

dproc96 (commented):
This looks really good, great job replicating the mocks! Only a couple minor changes. Also a thought is that if we're giving them the option to sign up with a password instead of Google, should we also provide them the option to log in with the password?

PR #38, opened 2021-01-14
Scheduler component

alvyjudy:
This is not completed yet but I would like to get some feedback before further drilling down on it. 
image


alvyjudy:
More changes have been made:
- selected text now has white color (thank Matt for the idea)
- data fetching hook to grab available time has been set up
- mechanism added to block unavailable date
- orange dots are now smaller

image

client/package.json
```diff
@@ -30,5 +30,11 @@
             "last 1 firefox version",
             "last 1 safari version"
         ]
+    },
+    "devDependencies": {
```
dproc96: These four should be standard dependencies because they'll be needed for deployment

dproc96 (commented):
I noticed some warnings in the console (some of them are unrelated to your PR so I'm not sure why those got through but we can address them later), can you fix those and the dependency issue? Otherwise this looks good to go

alvyjudy:
I have fixed the dependency issue and warnings. There are a few that still linger which I'm not sure where they originated. I will ask the team during our stand up.

dproc96 (approved):
This looks good to go, nice work! For that one remaining warning, I saw that Taras has already fixed it in his PR so that's fine to leave

PR #39, opened 2021-01-14
Small corrections to App.js

mattchx:
Corrected the spelling of component and added LandingPage as a Route


dproc96 (approved):
Good catch

PR #40, opened 2021-01-14
Google auth

kozaktar:
Features:

-Added user schema.
-Implemented google OAuth login.
-After successful login/signup with google account google id_token [used for account authentication] and access_token [used to access api] will be stored in cookies with corresponding names in the following location : "/"
-Added authentication controller to the backend server
-Added auth middleware to the backend:
   -auth middleware expects Authorization bearer token in the request header containing google  token

Potential issues:

-tokens expire in 1 hour at which point user will have to re login.


client/src/utils/googleAuth.js
```diff
@@ -0,0 +1,13 @@
+import axios from "axios";
+
+const userAPIpath = "http://localhost:3001/api/authentication/google";
```
dproc96, line 3:
```suggestion
const userAPIpath = "/api/authentication/google";
```
It looks like this is your problem. If you also change the proxy to "http:" instead of "https:" in the package.json then you should be good to go

mattchx: @kozaktar  Did this work?

kozaktar: No I was getting an error.  I used fetch and that seems to work

mattchx (commented):
Looks good! I've tested it and the google auth does work. Glad to see Fetch worked.

server/routes/controllers/authentication.js
```diff
@@ -0,0 +1,68 @@
...
+);
+
+// login with googleAuth
+router.post("/api/authentication/google", async (req, res) => {
+  //code from google auth, email=user email, variant=type of authentication [login vs signup]
+  const { code, email, variant } = req.body;
+  oAuthClient.getToken(code, async (err, token) => {
+    if (err) {
+      console.log(err);
+      res.status(500).send("Authentication error.", err);
```
dproc96, line 23: This line is crashing the app if there's an error. Express interprets res.send(a, b) as a being the status code and b being the body, even if status code is already set.

dproc96 (commented):
This looks good, just caught one error that is crashing the server when something goes wrong but it should be an easy fix. Once that's fixed, this should be good to go

server/routes/controllers/authentication.js
```diff
@@ -0,0 +1,68 @@
...
+      const userInfo = await oAuthClient.verifyIdToken({
+        idToken: tokens.id_token,
+        audience:
+          "294753578980-nbeunl8bovad0pp6t4ve8p5vso2hiahg.apps.googleusercontent.com"
+      });
+
+      const newUser = new User({
+        email: userInfo.payload.email,
+        name: userInfo.payload.name,
+        googleCredentials: id_token
```
alvyjudy, line 42: I think this is the reason the reason causing the token to expire after half an hour. The id_token is not the refresh token (see this [post](https://medium.com/@nilasini/id-token-vs-access-token-17e7dd622084)). I think it is the `code` sent from the client (in the `GoogleLoginButton.js`) that is the refresh token.

alvyjudy (commented):
I might have a hint on why the token expires after half an hour. See the comment on the relevant line

PR #41, opened 2021-01-15
Feature/event dialog

### LINKED ISSUE #7: FE: New Event Type Dialog
New dialog that allows you to create a new event type for a specific duration

WhoaNellie:
Bug where dialog blocks out the dashboard, and I need to remove the files I used to develop the integration code.
image
image

Edit: Fixed the dialog and grid issues!


client/src/pages/CreateEvent.js
```diff
@@ -0,0 +1,227 @@
...
+        </>
+    );
+}
+
+function MeetingCard({ type }) {
+    function addAppointment() {
+        axios
+            .post("/api/appointment", {
+                // dummy data
+                meeting_id: new mongoose.Types.ObjectId(),
```
dproc96: I don't think this line is needed and you can avoid importing mongoose on the frontend. If you want to pass the id for the meeting type, that can be passed as a string from the front end and the backend can use that to link to the meeting type `mongoose.Types.ObjectId(string)`

dproc96 (commented):
So far so good, aside from the changes you already noted. Also, there are some warnings in the console but I assume you'll resolve those as part of finishing this up. One other note, I think your screen size may be larger than mine because this is how the events appear on mine.

Screen Shot 2021-01-15 at 8 57 40 AM

client/src/pages/CreateEvent.js
WhoaNellie: I had to generate them to get the call to work because they have to be unique ids or the post will fail, I can leave it as non-working until we decide on how to generate meeting ids on the backend

PR #46, opened 2021-01-18
Secured front ent routes

kozaktar:
Added security to all front end routes:
-The only routes accessible to user who hasn't logged in are /login and /signup
-The other routes are only available once user has logged in.
-If user is logged in they can't access /login or /signup routes


client/src/App.js
```diff
@@ -10,25 +10,65 @@ import OnBoardingPage from "./pages/OnBoarding/OnBoarding";
...
-                <Route path="/home" component={Home} />
-            </BrowserRouter>
-        </MuiThemeProvider>
-    );
+  const [authenticated, setAuthenticated] = useState(false);
+
+  useEffect(async () => {
+    try {
+      const resp = await axios.get(
+        "http://localhost:3001/api/authentication/test",
```
sina-jamshidi: we should be able to use relative routes (eg. `/api/authentication/test`)

client/src/App.js
```diff
@@ -10,25 +10,65 @@ import OnBoardingPage from "./pages/OnBoarding/OnBoarding";
...
+      <BrowserRouter>
+        <Route exact path="/">
+          <Redirect to="/signup" />
+        </Route>
+        <Route
+          path="/signup"
+          render={() =>
+            authenticated ? <Redirect to="/home" /> : <SignUpPage />
+          }
+        />
```
sina-jamshidi: we can abstract this out to a "PrivateRoute" component. It will take the `SignUpPage` component as a prop, do the auth check, and display the component if authenticated.
[This is the general idea](https://medium.com/@tomlarge/private-routes-with-react-router-dom-28e9f40c7146)

sina-jamshidi (approved):
couple of small changes. Then you're good to merge!

Issue #8, opened 2021-01-06
Setup: Stripe API


sina-jamshidi:
https://stripe.com/docs/api/customers/create

https://stripe.com/docs/payments/checkout/subscriptions

Set up a dev account and use the appropriate API keys set up for the app.

PR #48, opened 2021-01-18
Feature/stripe integration

alvyjudy:
Ticket number: #8 
Here's how I intend to integrate stripe API into our app
image


client/src/App.js
```diff
@@ -10,25 +10,27 @@ import OnBoardingPage from "./pages/OnBoarding/OnBoarding";
...
-            </BrowserRouter>
-        </MuiThemeProvider>
-    );
+  return (
+    <MuiThemeProvider theme={theme}>
+      <BrowserRouter>
+        <Route exact path="/">
+          <Redirect to="/signup" />
+        </Route>
+        <Route path="/signup" component={SignUpPage} />
```
sina-jamshidi, line 24: this route is horrible

alvyjudy:
Thought more about Sina's comment, it does make more sense to only create the membership after the payment is completed. 
image

alvyjudy:
Current progress:
input
pending

failure
success




To do before merging: 
- Come up with sensible way of storing the payment information and match it with the logged in user
- Move the checkout component into <PrivateRoute/> but at the moment some errors are blocking me from doing so

client/src/components/CheckoutForm.js
```diff
@@ -38,6 +41,18 @@ const CheckoutForm = () => {
             </div>
           </Grid>
         </Grid>
+
+        <Grid item>
+          <Button
+            color="primary"
+            variant="contained"
+            className={classes.button}
+          >
```
kozaktar: this will redirect you to home because you're logged in.
Will cause a redirection "/"->"/signup"->"/home".

I think it should be  <Link to="/checkout" className={classes.link}>

kozaktar (changes requested):

client/src/components/Checkout/CheckoutSuccess.js
```diff
@@ -0,0 +1,74 @@
...
+        container
+        className={classes.grid}
+        direction="column"
+        alignItems="center"
+        justify="space-between"
+        spacing={4}
+      >
+        <Grid item>
+          <Typography variant="h5">
+            Your payment of {amount}$CAD is completed
```
mattchx: Is it possible to choose the currency?

server/routes/checkout.js
```diff
@@ -0,0 +1,14 @@
+const router = require("express").Router();
+const stripe = require("stripe").Stripe(process.env.STRIPE_SECRET);
+
+router.get("/api/checkout", async (req, res) => {
+  const amount = 120;
```
mattchx: This is for one time payment correct?

client/src/components/Checkout/CheckoutSuccess.js
alvyjudy: Yes but i haven't thought that far yet lol

server/routes/checkout.js
alvyjudy: I would need to do more work to resolve things like checking whether user has paid before

alvyjudy:
The stripe integration is ready! Next step is to strap it with the upgrade page. This is quite a big PR so I will break it down below. 

# Backend

Inside backend, I defined the Subscription.js mongoose model and a Subscription router. Inside the router file, there are two additional routes for test and development purposes. I want to hold on to those for now because it is likely we would be refactoring it.

The router exposes the following routes:

### GET `/api/subscription/check`

This route checks for whether the user has subscribed or not. If the user is subscribed, the response would send `{subscribed: true}`; otherwise, `{subscribed:false}`

### GET `/api/subscription/pay`

Starts a payment intent and send back the transaction ID and amount: `{clientSecret, amount}`

There is room for more options but I want to hold on to those until I integrate this with the upgrade page. 

I intentionally keep this route naive with regards to repeated payment and user authentication because its only duty is to bootstrap stripeAPI and it should be the frontend's duty to run the pre and post payment check and make sure the correct request is sent (personal take, but open for discussion 😜)

### POST `/api/subscription/set`

Add a document into the `Subscription` model with the provided client secret. For our current purpose, the presence of a document in this collection indicate that the user has paid and subscribed. We can extend this further in the future.

# Frontend

When the `Checkout.js` page is loaded, the first effect hook checks for whether the user is subscribed. If so, display a message and block the user from continuing the payment. If not, send a request to initiate a payment session and obtain the price and clientSecret.

Upon successful payment, redirect the user to the success page. Here I saved the reference value and amount in the URI so this page could be refreshed without losing information.

Upon failure, display a message and let user pay again.

# Testing

To do a successful payment, enter this credit card number: 4242 4242 4242 4242 with any expiry date and CVC (expiry can't be older than today)

alvyjudy:
Next step would be to convert manual payment into subscription model.

sina-jamshidi (commented):
Nice PR! Just look into using a subscription as discussed during standup, but otherwise looks good!

alvyjudy:
Pushed a route to handle canceling subscription

### GET `/api/subscription/cancel`

PR #49, opened 2021-01-18
Cleanup app.js

alvyjudy:
This PR

- cleans up the `app.js` on the server side so we have consistent import and route mounting. 
- move routes out of the controllers folder and fix the relative import paths
- extract the utility route handlers from the boilterplate and place them in `utils` folder
- remove boilerplates route


server/utils/fourOfourHandler.js
```diff
@@ -0,0 +1,7 @@
+const createError = require("http-errors");
```
WhoaNellie, line 1: Just want to say I like the name of this file haha

WhoaNellie (approved):

mattchx (approved):
Thanks for doing this DI !!

PR #50, opened 2021-01-19
Feature/fs events

### LINKED ISSUE #42: BE: Flesh out event types
- add event type title and unique link, colour name to schema
- integrate with event type dialog

WhoaNellie:
New event type schema and dialog updated and integrated!
image
image

Attempted to remove the warnings I thought were caused by my code, kind of hard to tell in some cases. I might wait to try to fix them after they're merged with the commits that fixed them a bit ago.


mattchx (approved):
You could change the Cancel button's color to a darker grey, or orange to make it stand out more. 
Also the mintues need a clock icon if we want to get closer to mocks. 
Other than that it looks great.

WhoaNellie:
> You could change the Cancel button's color to a darker grey, or orange to make it stand out more.
> Also the mintues need a clock icon if we want to get closer to mocks.
> Other than that it looks great.

There's no mocks for this part of the app, I was aiming to replicate Calendly's new event type UI, so I guess the closest would be this section
image
I decided to nix the 15/30/60 radio buttons as it was getting a little tight on space to fit this all in one dialog

Also can def make the cancel button a darker grey 👍

PR #54, opened 2021-01-19
Updated token

kozaktar:
-Updated JWT token to include user id and email
-added util function to decode token

Token structure example:

Jwt {
  header: JwtHeader { typ: 'JWT', alg: 'HS256' },
  body: JwtBody {
    user_id: '60073b42110ac1321c6b18b5',
    email: 'email@gmail.com',
    jti: '33631c20-273a-428c-990a-f4bf8a3bb1bb',
    iat: 1611086658,
    exp: 1611173058
  },

extract token using:

const jwtToken=req.cookies.app_auth_token;

decode using utils/jwtToken getTokenBody(jwtToken).

output:
{
  user_id: '60073b42110ac1321c6b18b5',
  email: 'email@gmail.com',
  jti: '33631c20-273a-428c-990a-f4bf8a3bb1bb',
  iat: 1611086658,
  exp: 1611173058
}

### DIFF
--- a/server/routes/controllers/authentication.js
+++ b/server/routes/controllers/authentication.js
@@ -47,8 +47,27 @@ router.post("/api/authentication/google", async (req, res) => {
       return;
     }
 
+    const app_user = await User.find({ email: userInfo.payload.email });
+    let claims = null;
+    console.log(app_user.length);
+    if (app_user.length > 0) {
+      claims = { user_id: app_user[0]._id, email };
+    } else {
+      console.log("creting....");
+      const newUser = new User({
+        email: userInfo.payload.email,
+        name: userInfo.payload.name
+      });
+      try {
+        await newUser.save();
+        claims = { user_id: newUser._id, email };
+      } catch (err) {
+        console.log(err);
+        res.status(400).send(err);
+        return;
+      }
+    }
     //create JWT token for route auth
-    const claims = { type: "route security", app: "calndApp" };
     const jwt_token = jwt.create(claims, process.env.JWT_SECRET);
     //token valid for 24h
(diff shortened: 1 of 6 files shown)


WhoaNellie (dismissed):
Looks good, thank you!!

WhoaNellie (approved):

alvyjudy (dismissed):
The change to the token looks good to me! 
Could you first merge the dev into your branch first (making sure nothing breaks upon resolving the conflict) before merging this PR?

Also noticed that the Authentication store isn't using the user id to hold the reference to User model. Might have to modify that down the road

server/utils/jwtToken.js
```diff
@@ -0,0 +1,9 @@
+
```
alvyjudy, line 1: I think this function could be converted into a middleware that adds the user Id to the request object and others can obtain it more easily.  Something like this:
```
const getUserId = (req, res, next)=>{
    req.userId = jwt(req.cookies.app_auth_token);
    next()
}
```
Instead of everyone adding code in the request handlers to extract the user id, having a user id available would be more convenient.

server/routes/controllers/authentication.js
```diff
@@ -47,8 +47,27 @@ router.post("/api/authentication/google", async (req, res) => {
       return;
     }
 
+    const app_user = await User.find({ email: userInfo.payload.email });
+    let claims = null;
+    console.log(app_user.length);
+    if (app_user.length > 0) {
+      claims = { user_id: app_user[0]._id, email };
```
mattchx, line 54: What does `claims` mean in this case?

mattchx (commented):
Nicely done!

PR #63, opened 2021-01-20
format code to use 2space indent

alvyjudy:
Just to keep things consistent


WhoaNellie (approved):
rip all the extra spaces I added 😢

PR #66, opened 2021-01-21
Bug fix on PrivateRoute: Defer rendering UI until authentication is confirmed

alvyjudy:
After some discussion with Taras, we worked out a new way to implement <PrivateRoute/> component. It solves the previous bug (not able to handle page refresh) by deferring the UI render until authentication is confirmed. User will be redirected to sign up page if authentication is invalid.


kozaktar:
can we redirect to login page instead?

sina-jamshidi (approved):

PR #76, opened 2021-01-24
Feature/ Photo upload

mattchx:
**Fully functional Photo Upload**
- image is previewed before upload
- saving changes posts the image to S3 bucket which returns a public url that is used in Avatar as `src`
- React Context is used to populate the image in the dashboard
- Material UI Dialog is used for upload modal


client/src/context/UserContext.js
```diff
@@ -0,0 +1,23 @@
+import React, { useState, createContext } from 'react';
+import PropTypes from 'prop-types';
+
+export const UserContext = createContext();
+
+export function UserContextProvider(props) {
+  const [url, setUrl] = useState('');
```
WhoaNellie: Might consider renaming this var to sometime like imgUrl/setImgUrl after we merge because there are some vars named url that deal with the users' unique urls. Depends on if they'll end up colliding on any pages though I suppose.

client/src/components/SelectTabs/EventTypesTab.js
```diff
@@ -267,7 +265,7 @@ export default function EventTypesTab() {
           <Box display="flex" className={classes.profileAndNewTypeBox}>
             <Avatar
               className={classes.avatar}
-              src={ProfileImage}
+              src=''
```
WhoaNellie, line 274: Is this meant to be src={url} or does this work somehow?

dproc96 (changes requested):
It looks like you haven't saved the image to the actual profile document, so it doesn't persist. Please see Nellie's comments as well. Also take a look at the console, I noticed quite a few errors when I ran the app so if those are related to this feature please address them before pushing.

client/src/context/UserContext.js
mattchx: Sure that's an easy fix.

client/src/components/SelectTabs/EventTypesTab.js
mattchx: React Context would not pass the updated url to `EventTypesTab.js` so I removed the var until we can get a fix.

mattchx:
I've addressed the concerns outlined by my reviewers.
Please take a look at my changes. Pay particular attention to my `db.User.updateOne()` function in the BE route.

* I have not corrected the FE compile errors as those are coming from other components unrelated to mine. I will merge dev into my branch when the latest version is approved and merged. I believe this will correct most/all of those errors. 

Thank you!

server/routes/imageUploaderAPI.js
```diff
@@ -0,0 +1,58 @@
...
+        db.User.updateOne({ _id: req.userId }, { $set: { photoUrl: AwsUrl } })
+          .then((response) => {
+            console.log(response)
+            res.send(response) 
+          })
+          .catch((error) => {
+            console.log(error);
+            res.status(500).send(error);
+          });
+      }
```
alvyjudy: When I ran the app, I'm getting an `Cannot set headers after they are sent to the client` error. I think these few lines might be why because you are attempting to send the response in `.then()` and `.catch()` after sending the response already. 

I suggest putting line 39-42 inside the `then` block. That way, you can guarantee that the url has been stored in the database before telling user that image is uploaded.

PR #78, opened 2021-01-25
Integrate calendar UI

kozaktar:
Integrated availability controller with calendar ui.

What's been done:

-Integrated availability controller with calendar.  Fee time slots are now populated from google calendar available slots.
-Fixed bug in availability controller where days with full availability (no busy slots would cause a 500 error)
-Fixed a bug where blocked off dates on the calendar would cause a pseudo race condition.
-Updated PickDate component to block off dates based on availability parameter [availability=available days set during onboarding].
-Updated PickTime component by adding interval parameter to set meeting length so that available time slots can be broken up based on meeting length.
-Updated PickTime component by adding availabilityTimes parameter [availabilityTimes= available times set during onboarding] allowing for available times from google calendar to be filtered based on availability times set during onboarding.

What needs to be done:

-Need to integrate onboarding data with calendar.


client/src/components/scheduler/PickTime.js
```diff
@@ -3,28 +3,34 @@ import { Grid, ListItem, makeStyles, Typography } from "@material-ui/core";
...
-      "19:30",
-      "20:00",
-    ]);
+    axios.get(`/api/calendar/availability?day=${isoDate}`, {
+      withCredentials: true
+    }).then(res=>{
+      setTimeSlots(getTimeSlots(res.data.availability, interval, availabilityTimes))
+      }).catch(err=>console.log(err))
+    
   }, [date]);
```
WhoaNellie, line 28: This isn't your line but I just noticed this. Since date isn't being set anywhere after its initialized is there any difference between this and just having this useEffect go onmount with `[]`?

client/src/pages/Scheduler.js
```diff
@@ -4,15 +4,30 @@ import {
...
+import { getNextAvailableDate } from "../utils/calendarUtils";
 
 const Scheduler = () => {
   const classes = useStyles();
+
+  //this will be fetched from the server
+  const availTimes={start:"09:00", end:"17:00"}
+  const availDates = [1, 2, 3, 4, 5];
+  //this will be set when picking event type --> pulled from context?
+  const interval="60";
```
WhoaNellie, line 23: Should this be based on the duration of the event type eventually?

client/src/utils/calendarUtils.js
```diff
@@ -0,0 +1,69 @@
...
+  const intervalLength = meetingLength * -1;
+
+  while (milliscondsToMinutes(meetingEndTime.diff(endTime)) <= intervalLength) {
+    const meetingStartHour = meetingEndTime.get("hours");
+    const meetingStartMinute = meetingEndTime.get("minutes");
+    const minutesFormated =
+      meetingStartMinute < 10
+        ? "0" + meetingStartMinute
+        : meetingStartMinute + "";
+    intervals.push(`${meetingStartHour}:${minutesFormated}`);
```
WhoaNellie, line 38: You might be able to use some date-fns methods to format the date easier since we're already importing it for other parts of the app
https://date-fns.org/v2.16.1/docs/formatDuration

client/src/utils/calendarUtils.js
```diff
@@ -0,0 +1,69 @@
+import moment from "moment";
```
WhoaNellie, line 1: We're using date-fns in other parts of the app for time formatting, might consider changing if there's nothing specific to moment you need

server/routes/googleCalendarAPI.js
```diff
@@ -15,9 +15,11 @@ router.get("/api/calendar/availability",auth, async (req, res) => {
     const tokenStore=await db.AuthStore.findOne({email:usr.email});
     const authToken=tokenStore.googleAuthToken;
     const availability= await getAvailability(authToken, day);
+    console.log(availability)
```
WhoaNellie, line 18: Can probably remove this

WhoaNellie (commented):
I like the calendarutils functions!

dproc96 (commented):
I would reiterate Nellie's comments, but otherwise this looks good. You should run prettier on your files though, there's some spacing and semicolon issues I saw

client/src/pages/Scheduler.js
kozaktar: yes, this is just dummy data. Eventually it should pull from event

client/src/components/scheduler/PickTime.js
kozaktar: you're right! [] would work just as well

PR #82, opened 2021-01-26
Add Global React context

mattchx:
**Completed**
- applied React context to pull user data from BE
- currently displaying user name and profile photo using context

**To Do**
- integrate Auth context with User context
- apply context to different parts of the app as required

**EDIT:** Global context for user data and auth is fully functional


alvyjudy:
I think we should place the provider in `App.js` instead because there are other components also needing access to the context data. That's also where other contexts are placed as well (MuiThemeProvider and BrowserRouter).

client/src/providers/Context.js
```diff
@@ -0,0 +1,52 @@
...
+};
+
+export const UserContextProvider = ({ children }) => {
+  const [userData, setUserData] = useState([]);
+
+  useEffect(() => {
+    axios
+      .get('/api/user/data')
+      .then((res) => {
+        console.log(res.data);
```
kozaktar: remove console.log after testing is done

client/src/components/Header/NavBar.js
```diff
@@ -39,6 +41,9 @@ export default function NavBar() {
   const classes = useStyles();
   const preventDefault = (event) => event.preventDefault();
 
+  const user = useUserContext()
+  const { name, photoUrl } = user
+
```
kozaktar, line 60: can be simplified to:
 const { name, photoUrl } = useUserContext();

kozaktar (commented):
Looks really good Matt!

client/src/providers/Context.js
mattchx: sure!

client/src/components/Header/NavBar.js
mattchx: Good point

mattchx:
> I think we should place the provider in `App.js` instead because there are other components also needing access to the context data. That's also where other contexts are placed as well (MuiThemeProvider and BrowserRouter).

Absolutely, I can update the draft.

mattchx:
Context is working. 
Currently issue where I cannot get context to update and rerender as user information is updated. 
I believe there are two options. 
1) To update the user state object directly by passing the updated variable (eg photoUrl) back into context. 
2) To trigger a new get request in context causing the user state object to be updated. (I plan to implement this)
- Bascially I have passed custom hooks that pass true back into context which then tiggers a useEffect to start a get request

Please let me know if you have ideas how to correct this or a better method. 
Thanks!

client/src/providers/Context.js
```diff
@@ -0,0 +1,71 @@
...
+// Set Auth
+export const useSetAuthenticated = () => {
+  // Usage:
+  // const setAuthenticated = useSetAuthenticated();
+  // setAuthenticated(true) OR setAuthenticated(false);
+  const setAuthenticated = useContext(UserContext).setAuthenticated;
+  return setAuthenticated;
+};
+// Get User Data
+export const useUserContext = () => {
```
alvyjudy: I think we should name this `useUserData` instead. 
In the comment let's also briefly outline what the data would include. Something like this:
```
// Usage: 
// const userData = useUserData(); 
// {_id, email, name, googleCredentials, photoUrl, URL, stripeId, subscribed, subscriptionId, ...}
```

which brings me to the question regarding how much of this data we want to expose to the frontend 🧐

client/src/providers/Context.js
```diff
@@ -0,0 +1,71 @@
...
+  return useContext(UserContext).userData;
+};
+// Set User Data has Updated
+export const useSetUserHasUpdated = () => {
```
alvyjudy: If I understand correctly with the set up, the child component calls a function which triggers the context to update on its own and renders the child with updated props. In that case, we should name this `useUpdateData` and have it return a function, which can be invoked without any argument.
```
const useUpdateData= ()=>{
    // Usage: 
    // const updateData = useUpdateData()
    // updateData()
    const setUserHasUpdated = useContext(UserContext).setUserHasUpdated;
    const updateData = () => {setUserHasUpdated(true)}
    return updateData
}
```

client/src/providers/Context.js
```diff
@@ -0,0 +1,71 @@
...
+
+  // Gets User Data from DB
+  useEffect(() => {
+    axios
+      .get("/api/user/data")
+      .then((res) => {
+        setUserData(res.data);
+      })
+      .catch((err) => console.log(err));
+  }, []);
```
alvyjudy, line 53: I'm not sure why this block is necessary. Wouldn't the second useEffect fire on page load anyways?

alvyjudy (commented):
Your code does seem to work for me actually. Could you elaborate on the issue you are experiencing? I left a few comments and suggestions based on my understanding of your set up. Let me know if they don't make sense because I might have misinterpreted your code. We can discuss them on a call tomorrow!

A big issue I noticed is too many re-rendering. I haven't come up with an idea where things could be optimized but worth looking into.

client/src/providers/Context.js
```diff
@@ -0,0 +1,71 @@
...
+
+  useEffect(() => {
+    axios
+      .get("/api/user/data")
+      .then((res) => {
+        setUserData(res.data);
+      })
+      .catch((err) => console.log(err));
+    setUserHasUpdated(false);
```
dproc96: The way you've written this, it will make the API request fire twice. The reason is because useEffect runs every time the value of its dependencies changes so it goes like this:

1. userUpdated is set to true
2. useEffect detects a change
3. The API request is fired
4. userUpdated is changed to false
5. useEffect detects a change
6. The API request is fired
7. userUpdated is set to false, but because it was already false the useEffect doesn't detect a change so it doesn't run a third time

server/routes/imageUploaderAPI.js
```diff
@@ -1,37 +1,54 @@
...
       } else {
-        res.status(200).send({msg:'Your image was uploaded sucessfully', fileName : req.file.key});
+        res.status(200).send({
+          msg: "Your image was uploaded sucessfully",
+          fileName: req.file.key,
+        });
+        const AwsUrl = `https://${BUCKET}.s3.amazonaws.com/${req.file.key}`;
+        db.User.updateOne({ _id: req.userId }, { $set: { photoUrl: AwsUrl } })
+          .then((response) => {
+            res.send(response);
```
dproc96: You're sending a response twice in this controller function which causes the app to crash

client/src/providers/Context.js
mattchx: That makes sense. I can make the name change.

client/src/providers/Context.js
mattchx: Yes, it does. I have removed the first useEffect.

client/src/providers/Context.js
mattchx: It's must cleaner without needing to be passed a boolean argument in the child component. Thanks!

server/routes/imageUploaderAPI.js
```diff
@@ -1,37 +1,54 @@
...
 
 // Set the region
 AWS.config.update({ region: process.env.AWS_REGION });
 
+const BUCKET = process.env.AWS_BUCKET_NAME;
+
 // Create S3 service object
 const s3 = new AWS.S3();
-const FN = Date.now().toString() + '.png'
+const FN = Date.now().toString() + ".png";
```
alvyjudy: This line is why image does not update!😜🧐

client/src/components/UploadModal/ImageUploader.js
```diff
@@ -88,22 +86,28 @@ export default function UploadDialog(props) {
...
-      const { fileName } = res.data;
-      setFilename(fileName);
-
+     
+      console.log(res.data.msg);
       setMessage(true);
-      await setTimeout(() => {
+      
+      // const {fileName} = res.data
+      // const newPhotoUrl = `https://${BUCKET}.s3.amazonaws.com/${fileName}`
```
sina-jamshidi: let's remove commented code and console.logs

server/routes/imageUploaderAPI.js
```diff
@@ -1,37 +1,55 @@
...
-        res.status(200).send({msg:'Your image was uploaded sucessfully', fileName : req.file.key});
+        res.status(200).send({
+          msg: "Your image was uploaded sucessfully",
+          fileName: req.file.key,
+        });
+        const AwsUrl = `https://${BUCKET}.s3.amazonaws.com/${req.file.key}`;
+        db.User.updateOne({ _id: req.userId }, { $set: { photoUrl: AwsUrl } })
+          .then((response) => {
+            res.send(response);
+          })
```
sina-jamshidi: 👍

sina-jamshidi:
> Context is working.
> Currently issue where I cannot get context to update and rerender as user information is updated.
> I believe there are two options.
> 
> 1. To update the user state object directly by passing the updated variable (eg photoUrl) back into context.
> 2. To trigger a new get request in context causing the user state object to be updated. (I plan to implement this)
> 
> * Bascially I have passed custom hooks that pass true back into context which then tiggers a useEffect to start a get request
> 
> Please let me know if you have ideas how to correct this or a better method.
> Thanks!

I'm a little worried that your way of fixing this is too hacky and technically a forced update, which is a React antipattern. Can you elaborate on the re-render issue? What is a specific case where it doesn't work?

server/routes/imageUploaderAPI.js
mattchx: Thanks for your help!

mattchx:
> > Context is working.
> > Currently issue where I cannot get context to update and rerender as user information is updated.
> > I believe there are two options.
> > 
> > 1. To update the user state object directly by passing the updated variable (eg photoUrl) back into context.
> > 2. To trigger a new get request in context causing the user state object to be updated. (I plan to implement this)
> > 
> > 
> > * Bascially I have passed custom hooks that pass true back into context which then tiggers a useEffect to start a get request
> > 
> > Please let me know if you have ideas how to correct this or a better method.
> > Thanks!
> 
> I'm a little worried that your way of fixing this is too hacky and technically a forced update, which is a React antipattern. Can you elaborate on the re-render issue? What is a specific case where it doesn't work?

The render was fixed by moving the profile image filename creation inside the upload function in the BE. 
You mean option 1 (see above) would be less hacky?

server/routes/imageUploaderAPI.js
```diff
@@ -42,13 +43,12 @@ router.post("/api/image-upload", auth, (req, res) => {
...
-            res.send(response);
-          })
-          .catch((error) => {
-            res.status(500).send(error);
-          });
+        db.User.updateOne(
+          { _id: req.userId },
+          { $set: { photoUrl: AwsUrl } }
+        ).catch((error) => {
+          res.send(error);
```
dproc96: I think this error handling would still crash the server if it came up, maybe waiting to send the response until after updating the DB instead of before would be a better way to handle it

dproc96 (commented):
One small thing I noticed in the API

server/routes/imageUploaderAPI.js
mattchx: Thanks for comments Dan! I did correct the error handling and also simplifed the logic in the request.

server/routes/imageUploaderAPI.js
mattchx: For some reason I thought I must send a response when the DB was updated. But then I realized that only one response is needed.

server/routes/imageUploaderAPI.js
```diff
@@ -1,37 +1,51 @@
...
       } else {
-        res.status(200).send({msg:'Your image was uploaded sucessfully', fileName : req.file.key});
+        const awsUrl = `https://${BUCKET}.s3.amazonaws.com/${req.file.key}`;
+        db.User.updateOne(
+          { _id: req.userId },
+          { $set: { photoUrl: awsUrl } }
+        ).catch((error) => {
+          res.send(error);
+        });
+        res.status(200).send({ awsUrl });
```
dproc96: You'll want to move this to a .then() on Line 45 to solve potential async issuea

mattchx: I did ` .then((res) =>  res.status(200).send({ awsUrl })` originally but caused the app to crash.
So now I've changed it to this: `.then(() => uploadRes.status(200).send({ awsUrl }))` which is working.

server/routes/imageUploaderAPI.js
```diff
@@ -1,37 +1,50 @@
...
     if (err) {
-      res.status(500).send('There was a server error with the upload');
+      res.status(500).send("There was a server error with the upload");
     } else {
       if (req.file == undefined) {
-        res.status(400).send('There was a error with your upload paramaters');
+        res.status(400).send("There was a error with your upload paramaters");
       } else {
-        res.status(200).send({msg:'Your image was uploaded sucessfully', fileName : req.file.key});
+        const uploadRes = res;
```
dproc96: I don't think it's necessary to rename the response here, the reason your original solution wasn't working was because adding `res` as an argument in the .then() reassigned the value

mattchx: That makes sense!

PR #84, opened 2021-01-26
Check link while typing

kozaktar:
Features:

-Updated link inputs on new event dialog and onboarding page to check for link uniqueness as the user is typing
-Added denouncing to check link api calls.
-Updated onboarding link input styles to be similar to new event dialog link input

https://user-images.githubusercontent.com/20387499/105861141-67821200-5fbc-11eb-9899-103f70403f12.mp4

https://user-images.githubusercontent.com/20387499/105861145-681aa880-5fbc-11eb-800a-561f96318b1d.mp4


client/src/utils/utils.js
```diff
@@ -0,0 +1,12 @@
+//author medium.com/@kartikag01
```
alvyjudy, line 1: If we are abstracting a utility function, we should properly name the file (like `debounceFunction.js`) and give it a simple comment outlining its usage and purpose.

kozaktar: I added comments to debounce function.  We may end up having more than one general utility function so I think I'll keep it in util file for now.

client/src/components/onboarding/SetTimezoneUrl.js
```diff
@@ -1,20 +1,51 @@
...
   const { url, setUrl, timezone, setTimezone } = props;
 
+  const [unique, setUnique] = useState(true);
+
+  
+
+  const checkUnique=async (linkVal)=>{
+    try {
+      const response = await axios.get(
+        `/api/user/is_unique?URL=${linkVal}`
```
mattchx: So if an error is returned then `setUnique = false` correct? Could the same thing be done where the get request returns a value of false?

client/src/utils/utils.js
mattchx: What does this function do?

mattchx (approved):
Nicely done !

client/src/utils/utils.js
kozaktar: This function takes a function as the first argument and time in milliseconds as second argument and returns a function that runs the original function (argument 1) a maximum of once in a period of time supplied as (argument 2).  Here is an article explaining debouncing in detail: https://levelup.gitconnected.com/debounce-in-javascript-improve-your-applications-performance-5b01855e086

client/src/components/onboarding/SetTimezoneUrl.js
kozaktar: That's correct.  It can be done as you've suggested, but I think catch statement works a little better here as it will set unique=false if the call fails for any reason.

client/src/components/SelectTabs/EventTypesTab.js
```diff
@@ -219,14 +237,42 @@ export default function EventTypesTab() {
...
+      setUnique(true);
+    } catch (err) {
+      console.log(err);
+      setUnique(false);
+    }
+  
+  }
+
+
+  const debounceCheckUnique= useCallback(debounce(checkUnique, 500), []);
```
WhoaNellie, line 269: Nice!

WhoaNellie (approved):

dproc96 (approved):
Nice work, clever to use the error to handle not unique because then you're covered if the call fails for any reason

PR #94, opened 2021-01-26
Removing console errors for Grid sizing/Spacing

WhoaNellie:
Passing in numbers instead of strings to resolve errors


kozaktar (approved):
Thanks for doing this! Console looks much better now!

PR #95, opened 2021-01-27
Feature/create a team UI

alvyjudy:
# Feature

- add a team tab and with a "create team button"
- "create team button" opens a dialog
- implement a list to specify the email to send invitation to

# Next step for enhancement (probably in another PR?)

- Enhance email validation (avoid duplication)
- Other features in the dialog?

image

image


dproc96 (commented):
It looks like you have to rename createTeam.js to CreateTeam.js to get it running

Also I find the positioning in the dialog kind of odd, the labels take up half of the space which makes it look a little unbalanced and makes it harder to use in mobile. Maybe try the label property for Text Field, that might look cleaner. Additionally the text is vertically centered in the description box which also looks a little odd.

Otherwise this is a great start!

alvyjudy:
Made some edits for the styling
image

client/src/components/teams/CreateTeam.js
```diff
@@ -0,0 +1,148 @@
...
+const CreateTeam = ({ open, closeDialog }) => {
+  const classes = useStyles();
+
+  const createTeam = (event) => {
+    event.preventDefault();
+  };
+
+  const [teamName, setTeamName] = useState("");
+  const handleTeamNameChange = (event) => {
+    event.preventDefault();
```
sina-jamshidi: `preventDefault` is not necessary in these cases of typing in a textfield. The main case for when you need it is when you have a `type="submit"` button

client/src/components/teams/CreateTeam.js
```diff
@@ -0,0 +1,148 @@
...
+} from "@material-ui/core";
+
+import InviteMembers from "./InviteMembers";
+import MembersToBeInvited from "./MembersToBeInvited";
+
+const CreateTeam = ({ open, closeDialog }) => {
+  const classes = useStyles();
+
+  const createTeam = (event) => {
+    event.preventDefault();
```
sina-jamshidi: not necessary here either. For now you can just console.log() or something

client/src/components/teams/InviteMembers.js
```diff
@@ -0,0 +1,73 @@
...
+  makeStyles,
+} from "@material-ui/core";
+
+const InviteMembers = ({ setTeamMembers, teamMembers }) => {
+  const classes = useStyles();
+
+  const [email, setEmail] = useState("");
+  const [emailValid, setEmailValid] = useState(false);
+  const handleEmailChange = (event) => {
+    event.preventDefault();
```
sina-jamshidi: don't need this

client/src/components/teams/InviteMembers.js
```diff
@@ -0,0 +1,73 @@
...
+          value={email}
+          onChange={handleEmailChange}
+          margin="normal"
+          placeholder="Email address"
+        />
+      </Grid>
+      <Grid item xs={2}>
+        <Button
+          onClick={addEmailToInvitedList}
+          disabled={!emailValid}
```
sina-jamshidi, line 49: nice UX detail!

client/src/components/teams/InviteMembers.js
```diff
@@ -0,0 +1,73 @@
...
+} from "@material-ui/core";
+
+const InviteMembers = ({ setTeamMembers, teamMembers }) => {
+  const classes = useStyles();
+
+  const [email, setEmail] = useState("");
+  const [emailValid, setEmailValid] = useState(false);
+  const handleEmailChange = (event) => {
+    event.preventDefault();
+    //validation could probably be done better
```
sina-jamshidi, line 17: if you can think of a better way you should do it 😅 
This looks fine for now. If we get to it we can make this an autocomplete box that queries the current users

client/src/components/teams/MembersToBeInvited.js
```diff
@@ -0,0 +1,67 @@
...
+import React from "react";
+
+const MembersToBeInvited = ({ teamMembers, setTeamMembers }) => {
+  const classes = useStyles();
+  const teamMemberNotEmpty = teamMembers.length !== 0;
+  const removeEmailFromInvitedList = (memberToRemove) => () => {
+    // Caveat: each email must be unique in the list for this to work
+    const filteredList = teamMembers.filter(
+      (member) => member.email !== memberToRemove.email
+    );
```
sina-jamshidi, line 18: Caveat makes sense. The way to handle that would be to just not allow adding duplicate emails, which is doable because you did a good job to organize the structure so both components are passed the `teamMembers` as a prop.

client/src/components/teams/MembersToBeInvited.js
```diff
@@ -0,0 +1,67 @@
...
+  };
+  return (
+    <Grid container direction="column" className={classes.root}>
+      <Grid item>
+        <InputLabel className={classes.label}>
+          An invitation email will be sent to the following members:
+        </InputLabel>
+      </Grid>
+      <Grid item container direction="column" wrap="nowrap">
+        {teamMemberNotEmpty &&
```
sina-jamshidi, line 29: If the teamMembers array is empty, the map doesn't display anything so there's no need for the check.
Also since `teamMemberNotEmpty` is used only once, we could've just done `teamMembers.length > 0 &&`

client/src/components/teams/MembersToBeInvited.js
```diff
@@ -0,0 +1,67 @@
...
+          teamMembers.map((member, index) => (
+            <Grid item container key={index} alignItems="center" spacing={3}>
```
sina-jamshidi: key should be something unique like the email to be safer. Otherwise if there's another map using index, or the order of the members changes, etc. It could cause re-render issues

client/src/components/teams/MembersToBeInvited.js
```diff
@@ -0,0 +1,67 @@
...
+              <Grid item xs={1}>
+                <Brightness1 color="primary" className={classes.icon} />
+              </Grid>
+              <Grid item xs={4}>
+                <Typography>{member.email}</Typography>
+              </Grid>
+              <Grid item xs={4}>
+                <Button onClick={removeEmailFromInvitedList(member)}>
```
sina-jamshidi, line 45: I personally prefer inline function `onClick={() => removeEmailFromInvitedList(member)}` in terms of readability to currying, which is your approach here, but both are valid.

sina-jamshidi (approved):
Looks good. A couple of small things to fix. This was a good component layout and how I would have done it. This dialog will take more shape as we get more backend functionality done as well.
Also UI wise, try to replicate Nellie's event type dialog for the sake of consistency, I think that will playing around with the grid sizings to get the spacing  to be closer together, with the textfields taking up more space than the labels

mattchx (approved):
Nice FE work!

client/src/components/teams/MembersToBeInvited.js
alvyjudy: If the array is empty I think we would get an error when doing `[].map(member=>{member.email})` that's why we included the length check, but inlining it makes sense!

client/src/components/teams/MembersToBeInvited.js
alvyjudy: This makes sense, I will use email instead.

alvyjudy:
More changes are made to the styles! Latest look:
image

dproc96 (approved):
This looks good to go!

PR #98, opened 2021-01-28
Create teamEventAPI.js

mattchx:
### Created 4 new routes 
1 - Create a new team event type
2 - Add or remove members from an event type
3 - Delete an event type
4 - Update an event type

For 2)  - Add or remove members from an event type
-- Could this be done in one request by passing in an updated member object?
Eg db.collection.update(query, updatedMemberObject)

**Edit:** Routes that update have been updated with `$set` and `$push` operators

**Edit:** TESTED AND WORKING!


server/routes/teamEventAPI.js
```diff
@@ -0,0 +1,66 @@
...
+const db = require("../db/models");
+const auth = require("../middleware/auth");
+
+const router = express.Router();
+
+// 1 - Create a new team event type
+router.post("/api/team-event", auth, (req, res) => {
+  if (req.body.duration) {
+    //
+    db.TeamEventType.create({ ...req.body, teamId: req.teamId })
```
dproc96: You should be able to keep this as EventType depending on Nellie's approach

server/routes/teamEventAPI.js
```diff
@@ -0,0 +1,66 @@
...
+        res.status(500).send(error);
+      });
+  } else if (!req.body.duration) {
+    res.status(400).send("Duration is required");
+  }
+});
+
+// 2 - Add or remove members from an event type
+// an updated team member is passed to DB to update added and removed members 
+router.post("/api/team/id/:id", (req, res) => {
```
dproc96:
```suggestion
router.post("/api/team/:id", (req, res) => {
```
You can probably just keep the id parameter

dproc96 (commented):
Good idea to use one route for adding or removing users, that can probably simplify the front end too!

server/routes/teamEventAPI.js
WhoaNellie: I think it could make sense to keep it as EventType, then change the EventType schema to include a teamMembers array or a teamId

server/routes/teamEventAPI.js
```diff
@@ -0,0 +1,66 @@
...
+  db.Team.update({ _id: req.params.id })
```
WhoaNellie: I think you'll need a second param here to update, it will probably be something like `db.Team.update({ _id: req.params.id }, { $push: { teamMembers: req.body.memberId}}` 
Just made up some variable names, but `$push` because I think teamMembers will be an array and memberId is on the req.body because I find that easier than doing params.
https://mongoosejs.com/docs/api.html#model_Model.update
$push and $set are atomic operators
https://docs.mongodb.com/manual/tutorial/model-data-for-atomic-operations/#pattern

server/routes/teamEventAPI.js
mattchx: Thanks Dan and Nellie! I will adjust the routes to your approach.

server/routes/teamEventAPI.js
```diff
@@ -0,0 +1,69 @@
...
+      res.send("Deleted team event type");
+    })
+    .catch((err) => {
+      res.status(400).send(err);
+    });
+});
+
+// 4 - Update an event type
+router.post("/api/team-event/:id", (req, res) => {
+  db.EventType.updateOne({ _id: req.params.id }, { $push: { $set: req.body } })
```
mattchx, line 57: I

server/routes/teamEventAPI.js
```diff
@@ -0,0 +1,69 @@
+const express = require("express");
+const db = require("../db/models");
+const auth = require("../middleware/auth");
+
+const router = express.Router();
+
+// 1 - Create a new team event type
+router.post("/api/team-event", auth, (req, res) => {
```
mattchx: If you have suggestions to change the route names please let me know. 
Also **auth** will be required for team routes, correct?

server/routes/teamEventAPI.js
```diff
@@ -0,0 +1,69 @@
...
+    res.status(400).send("Duration is required");
+  }
+});
+
+// 2 - Add or remove members from an event type
+// an updated team member object is passed to DB to update added and removed members
+router.post("/api/team/:id", (req, res) => {
+  db.Team.updateOne(
+    { _id: req.params.id },
+    { $push: { teamMembers: req.body.memberId } }
```
mattchx: `$push` allows new members to be added.
But can this handle deleting new members in this case? 
Would it be a good idea to remove the `$push` as this would force overwriting the team members object?

server/routes/teamEventAPI.js
```diff
@@ -0,0 +1,69 @@
...
+      res.send("Deleted team event type");
+    })
+    .catch((err) => {
+      res.status(400).send(err);
+    });
+});
+
+// 4 - Update an event type
+router.post("/api/team-event/:id", (req, res) => {
+  db.EventType.updateOne({ _id: req.params.id }, { $set: req.body })
```
mattchx: The EventType is updated using `{ $set: req.body }`

server/routes/teamEventAPI.js
WhoaNellie: Hmm, I didn't take into account that this route is for adding and removing. In that I suppose you wouldn't use $push and possibly the FE would be in charge of creating the new array they want to send over

server/routes/teamEventAPI.js
```diff
@@ -0,0 +1,69 @@
...
+const db = require("../db/models");
+const auth = require("../middleware/auth");
+
+const router = express.Router();
+
+// 1 - Create a new team event type
+router.post("/api/team-event", auth, (req, res) => {
+  if (req.body.duration) {
+    //
+    db.EventType.create({ ...req.body, teamId: req.teamId })
```
dproc96: @WhoaNellie Have you added teamId to the EventType model in your PR? If not then Matt should add it in this one.

Also wouldn't the teamId be a part of the req.body?

server/routes/teamEventAPI.js
```diff
@@ -0,0 +1,69 @@
...
+      }
+      res.send("Deleted team event type");
+    })
+    .catch((err) => {
+      res.status(400).send(err);
+    });
+});
+
+// 4 - Update an event type
+router.post("/api/team-event/:id", (req, res) => {
```
dproc96: This should probably be a put request to match conventions, it may also cause conflicting routes issues with the create route if they're both post requests (but I'm not positive in this case)

server/routes/teamEventAPI.js
WhoaNellie: I did add it yes

server/routes/teamEventAPI.js
mattchx: I will change the 2nd route(Add or remove members from an event type) to a PUT since I've read that it can be used for overwriting data.

server/routes/teamEventAPI.js
```diff
@@ -0,0 +1,70 @@
...
+      });
+  } else if (!req.body.duration) {
+    res.status(400).send("Duration is required");
+  }
+});
+
+// 2 - Add or remove members from an event type
+// an updated team member object is passed to DB to update added and removed members
+router.put("/api/team/:id", (req, res) => {
+  db.Team.updateOne(
```
mattchx: I found out on [stack overflow](https://stackoverflow.com/questions/38405989/mongoose-findoneandupdate-with-set-flag) that the $set operator use is implied whether not it's explicitly written. The function of $set is to force a merge with the current data. 
We do need the { overwrite: true }  because that would force an overwrite of data.
eg: `WorkPlan.findOneAndUpdate({ _id: req.params.id }, updateObj, { overwrite: true }, function(err) { ...})`

server/routes/teamEventAPI.js
```diff
@@ -0,0 +1,70 @@
...
+        res.status(500).send(error);
+      });
+  } else if (!req.body.duration) {
+    res.status(400).send("Duration is required");
+  }
+});
+
+// 2 - Add or remove members from an event type
+// an updated team member object is passed to DB to update added and removed members
+router.put("/api/team/:id", (req, res) => {
```
mattchx: In this case would it make more sense to search by team name instead?

server/routes/teamEventAPI.js
```diff
@@ -0,0 +1,70 @@
+const express = require("express");
+const db = require("../db/models");
+const auth = require("../middleware/auth");
+
+const router = express.Router();
+
+// 1 - Create a new team event type
+router.post("/api/team-event", auth, (req, res) => {
+  if (req.body.duration) {
```
kozaktar: Just wondering why we are checking for duration attribute? Is duration the only attribute that's required to create a team event? Also should we make duration required in the schema then we don't have to check for it.  If not supplied creating new team event will fail, and it will be caught by the catch block.

kozaktar (approved):
Really good work Matt. The code looks very clean.

server/routes/teamEventAPI.js
mattchx: In eventAPI, it requires the duration attribute so I followed the same patterned here.  I agree there may be ways to optimize this. Once we start integrating with FE I will get a better idea. But for now I'm open to suggestions 👍🏼

server/routes/teamEventAPI.js
dproc96: Probably not since we don't know if team name is unique

mattchx:
I did leave in some console.logs in BE routes so we can understand what the error checking is doing

PR #99, opened 2021-01-28
Manage your team UI

kozaktar:
Manage your team UI:

-Active users tab:
contains a list of active users with the following fields: name, role, calendarApp page, date added.  For each user you can remove user or change role.

-Pending tab:
contains a list of users who have been invited to join the team but haven't accepted yet.

-Team events tab:
Contains a list of team events. You can create new team event using new team event button.

-New team event dialog:
A modified version of new event type dialog that also contains a field for team members to invite.

-New user button allows you to invite new users to your team

-Team name can be modified by clicking on the name which will call up team name input field.

https://user-images.githubusercontent.com/20387499/106465947-ad315580-6468-11eb-9f7b-ddce8fb881f0.mp4


dproc96:
This looks good so far. I think yeah, see if you can accomplish it with the same component but if that doesn't work you can create a new one. For adding a new user, I think you can keep it pretty simple for now at least, maybe like Google Docs for reference

client/src/components/UserManagment/TeamNameInput.js
```diff
@@ -0,0 +1,66 @@
...
+          onClick={handleSubmit}
+          classes={{ root: classes.buttonCheck }}
+        >
+          <CheckIcon />
+        </IconButton>
+        <IconButton
+          color="primary"
+          aria-label="decline"
+          component="span"
+          onClick={toggleNameChange}
```
dproc96: As a user I would expect this button to undo my changes

dproc96 (commented):
This looks pretty good, had one odd UX question. Also do we want to keep the templates tab for the time being in case we have time to implement or remove it?

PR #100, opened 2021-01-28
Feature/reschedule

alvyjudy:
After confirming an appointment, user can visit `/appointment/reschedule/:appointmentId` or `/appointment/cancel/:appointmentId` to reschedule or cancel the appointment. 

# Frontend

## Routes added in `App.js`

- `/appointment/reschedule/:appointmentId`
- `/appointment/cancel/:appointmentId`
- `/appointment/cancelled`
- `/appointment/does-not-exist`

## Components

`RescheduleOrCancelAppointmentPage.js`:

Display the page for cancelling or rescheduling appointment. Redirect user to success page, error page or scheduling page

`RescheduleOrCancelAppointment.js`:

Display the input box and button for sending request to cancel appointment

`AppointmentDoesNotExistOrCancelled.js`:

Display the success/error page

## Implementation details

Components uses a `variant` prop to decide whether to render a cancel or reschedule page. The value is either `reschedule` or `cancel`. 

## Changes that might break other branches

`Scheduler.js` uses `useParams()` hook instead of regex to grab the host name and event name. Correspondingly, the path to scheduler is changed to `/appt/:hostName/:eventName` (previously `/appt/`)

# Backend

## Routes added

- `/api/appointment/cancel/:appointmentId`
- `/api/appointment/detail/:appointmentId`

## Middleware 

- `ensureAppointmentExists.js` (mounted to the above two routes)

# UI

Since I have not been able to fully test the feature (due to not being able to get availability on my end), I will include the UI screenshots here for now. 

## Cancel page:
Screen Shot 2021-01-28 at 6 34 31 PM

## Reschedule page
Screen Shot 2021-01-28 at 6 34 44 PM

## appointment does not exist page
image

## appointment successfully cancelled page
image


server/middleware/ensureAppointmentExists.js
```diff
@@ -0,0 +1,26 @@
...
+  // Otherwise, send a 400 response
+  //
+  // URL must contain:  `/:appointmentId`
+  // Example:
+  // `/api/appointment/:appointmentId`
+  // `/api/appointment/cancel/:appointmentId`
+  // `/api/appointment/reschedule/:appointmentId`
+
+  const { appointmentId } = req.params;
+  const result = await dbAppointment.findOne({ _id: appointmentId });
```
dproc96: If you're finding the appointment in the middleware, it doesn't make sense to find it again in the actual controller. You could just add this to the request before moving on

client/src/components/scheduler/RescheduleCancelAppointment.js
```diff
@@ -0,0 +1,72 @@
...
+  const reschedule = variant === "reschedule";
+  const cancel = variant === "cancel";
+  const classes = useStyles();
+  const cancelAppointment = (event) => {
+    axios.delete(`/api/appointment/cancel/${appointmentId}`).then(() => {
+      setAppointmentCancelled(true);
+    });
+  };
+
+  const linkToReschedule = `/appt/${eventUrl}`;
```
dproc96: Should the host url be included in here?

client/src/pages/reschedule/RescheduleOrCancelAppointmentPage.js
```diff
@@ -0,0 +1,78 @@
...
+      })
+      .catch(() => {
+        setAppointmentDoesNotExist(true);
+      });
+  }, [appointmentId]);
+
+  return (
+    <>
+      {appointmentCancelled && <Redirect to="/appointment/cancelled" />}
+      {appointmentDoesNotExist && <Redirect to="/appointment/does-not-exist" />}
```
WhoaNellie, line 41: Could this instead redirect to the 404 page taras is working on?

alvyjudy: This page actually has another purpose, if the user try to return to the page after cancelling, it should tell them that the appointment has been cancelled rather than sending them back to an unrelated error page

client/src/components/scheduler/RescheduleCancelAppointment.js
alvyjudy: Since I'm reusing the scheduler component, the url is required. After clicking on the reschedule button, the frontend sends a request to the backend to cancel the appointment, and redirects the user to the scheduling page, which requires the above link to work (not the most elegant solution but Sina and Nellie suggested to go with the simple way for now.)

server/middleware/ensureAppointmentExists.js
alvyjudy: That's true. I'll make the edit!

client/src/components/scheduler/RescheduleCancelAppointment.js
```diff
@@ -0,0 +1,72 @@
...
+import { Link } from "react-router-dom";
+import axios from "axios";
+
+const RescheduleCancelAppointment = ({ variant, appointmentId, eventUrl }) => {
+  const reschedule = variant === "reschedule";
+  const cancel = variant === "cancel";
+  const classes = useStyles();
+  const cancelAppointment = (event) => {
+    axios.delete(`/api/appointment/cancel/${appointmentId}`).then(() => {
+      setAppointmentCancelled(true);
```
dproc96: You haven't destructured this with the rest of the props. That being said, this line is also unnecessary because the component will be unmounted before this actually runs, which will throw an error.

Screen Shot 2021-01-29 at 8 55 11 AM

client/src/components/scheduler/RescheduleCancelAppointment.js
dproc96: Oh I see, that contains the host url as well. I thought it was just the specific event type url

client/src/components/scheduler/RescheduleCancelAppointment.js
```diff
@@ -0,0 +1,72 @@
...
+              multiline
+              rows={7}
+              variant="outlined"
+              fullWidth
+            />
+          )}
+        </Grid>
+        <Grid item>
+          <Button className={classes.button} onClick={cancelAppointment}>
+            <Link
```
dproc96: I wonder if linking them to reschedule at the same time as sending the request means that the availability shown will assume this meeting is still happening. I also wonder if that matters since they're rescheduling anyway

client/src/pages/Scheduler.js
```diff
@@ -47,12 +49,12 @@ const Scheduler = () => {
...
   useEffect(() => {
-    const regex = /\/.*\/.*\//g;
-    let searchUrl = path.match(regex)[0].slice(6);
-    const queryURL = `/api/event_details/${searchUrl}`;
+    console.log("name", hostName, eventName);
+    const queryURL = `/api/event_details/${hostName}/${eventName}`;
     axios.get(queryURL).then((res) => {
       // TODO: redirect to 404 page on no event found?
       let event = res.data[0];
+      console.log(event);
```
dproc96: You can remove this

client/src/pages/Scheduler.js
```diff
@@ -47,12 +49,12 @@ const Scheduler = () => {
   });
 
   useEffect(() => {
-    const regex = /\/.*\/.*\//g;
-    let searchUrl = path.match(regex)[0].slice(6);
-    const queryURL = `/api/event_details/${searchUrl}`;
+    console.log("name", hostName, eventName);
```
dproc96: This can be removed

dproc96 (commented):
Noticed a blocker and a couple of small things, overall the flow of it looks good though

client/src/components/scheduler/RescheduleCancelAppointment.js
alvyjudy: Right that's something I haven't considered. This downside sounds acceptable to me though since we are choosing to simplify the rescheduling flow. Can definitely discuss this during stand up!

client/src/components/scheduler/RescheduleCancelAppointment.js
alvyjudy: Right that's a bug. I might have put it there in the first place trying to use the parent component to do the redirection. Actually, that would solve the issue of cancelling and redirecting at the same time!

client/src/components/scheduler/RescheduleCancelAppointment.js
alvyjudy: update: This is actually resolved after I let the parent component handle redirection.

alvyjudy:
I have made the edit that solves the issue of simultaneously canceling the appointment and scheduling (which may block the timeslot) by having the parent component handle all redirection, alongside other small fixes.

Still can't get my availability to work. It seems like Taras will be busy so I will dig into the code and see if I can get it to work

dproc96 (approved):
Really good work accomplishing this and minimizing added code!

PR #101, opened 2021-01-28
Team model

WhoaNellie:
Starting to think about the team model, I think it could be as simple as this if we modify the User model to have their role and date added. Right now, I'm planning on having `members` be an array of users we can populate when working with the data and sorting it for the Round Robin scheduling. 

- Calendly's description of their round robin distributed events: "If someone has been assigned too many meetings of this type, their availability will be hidden from the scheduling page until others catch up." -- Should we hide availability of users if they've been assigned too many meetings? And how do we determine what "too many" is?

### DIFF
--- a/server/db/models/Team.js
+++ b/server/db/models/Team.js
@@ -1,0 +1,17 @@
+const mongoose = require("mongoose");
+const { User } = require("./User");
+
+const Schema = mongoose.Schema;
+
+const schema = new Schema({
+  members: {
+    type: [User],
+  },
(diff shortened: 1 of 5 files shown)


server/db/models/Team.js
```diff
@@ -0,0 +1,17 @@
+const mongoose = require("mongoose");
```
sina-jamshidi, line 1: - event types having a `members` array
- user model have `isAdmin` flag, `teamId` ref

PR #103, opened 2021-01-29
Handle invalid page urls in FE Router

mattchx:
### Completed
- 404 page is styled
- App.js contains /404 route for PageNotFound component
- "/" redirect to home page
### To do 
Currently redirects force the app to be locked to one page.
Eg:` <Redirect from="*" to="/404" />` and `<Route path="*" component{PageNotFound} />`  both lock the page to 404. 
I believe the issue may have something to do the the logic in public/private routes.


client/src/App.js
```diff
@@ -59,6 +60,11 @@ function App() {
         <PrivateRoute path="/checkout">
           <Checkout />
         </PrivateRoute>
+        <Route path="/404">
+          <PageNotFound />
+        </Route>
+        <Route exact path="/" component={Home} />
+        {/* <Redirect from="*" to="/404" /> */}
```
WhoaNellie, line 67:
```suggestion
        <Route component={PageNotFound} />
```
This seems to work for me! The wildcard * in the from/to will match all routes, but I think this will just match everything else.

Found in [this article](https://ultimatecourses.com/blog/react-router-not-found-component) although we're not using switch here

WhoaNellie:
```suggestion
        <Route component={PageNotFound} />
```

This route and wrapping all the routes in a Switch component worked for me!
https://ultimatecourses.com/blog/react-router-not-found-component

alvyjudy: I would recommend using a Switch component too. It renders the last item if no match is found. The caveat is that it only renders the first matching item as well but for our usage that shouldn't post an issue

client/src/pages/PageNotFound.js
```diff
@@ -0,0 +1,94 @@
...
+        </Toolbar>
+      </AppBar>
+      <Grid
+        className={classes.root}
+        container
+        justify="center"
+        alignItems="center"
+        spacing={0}
+      >
+        <Grid item className={classes.imageBox}>
```
dproc96:
```suggestion
        <Grid sm={12} md={4} item className={classes.imageBox}>
```
Using grid sizing like this will make the page display properly on medium sized screens (like mine)

client/src/App.js
mattchx: I considered that before but was concerned the switch component conflicts with the browser router. That's not the case?

WhoaNellie: I don't believe so, they show an example of them both in the docs!
https://reactrouter.com/web/guides/quick-start

PR #107, opened 2021-01-31
Feature/ Toggle Event Types On/Off

mattchx:
**This feature is fully functional!** 
_Here are the details:_
1. styled Event Cards to create an on/off switch which makes card link disabled
2. added `active` (boolean) to EventType schema
3. created put request inside `<EventCard>` which sends boolean value and the event's unqiue id
4. create a new route in `eventAPI.js` which updates `active` to true or false in DB
5.  designed `<EventNotActivePage>`
6.  invalid page is displayed based on the active status of the event inside the `<Scheduler>` page

In addition: 
I may make refinements to the styling. But overall, I am very happy with the result. 
Please let me know if you have any suggestions!


client/src/components/UserEvents/EventCard.js
```diff
@@ -25,13 +33,23 @@ export default function EventCard({ name, duration, color, link, url }) {
       marginBottom: 12,
     },
     colorBar: {
-      background: color || deepOrange[500],
+      background: !eventActive ? "lightgrey" : color || deepOrange[500],
```
WhoaNellie, line 36: 😃

client/src/components/SelectTabs/EventTypesTab.js
```diff
@@ -238,6 +238,7 @@ export default function EventTypesTab() {
         setUserURL(res.data);
       })
       .catch((err) => console.log(err));
+      console.log(userEvents)
   }, []);
```
WhoaNellie: If you could go through and make sure there are no console warnings from your code after you get the changes from my merge before you merge that would be great!

client/src/pages/Scheduler.js
```diff
@@ -64,95 +67,100 @@ const Scheduler = () => {
         duration: event.duration,
         link: event.link,
       });
+
       setAppointmentDetails({ ...appointmentDetails, eventId: event._id });
+      setEventActive(event.active);
```
WhoaNellie: Could you use the eventDetails state and add an active key to it?

client/src/components/UserEvents/EventCard.js
```diff
@@ -41,24 +59,56 @@ export default function EventCard({ name, duration, color, link, url }) {
...
 
+  const handleSwitch = async () => {
+    axios
+      .put("/api/event/toggle-active", {
+        active: !eventActive,
+        eventId: _id,
+      })
+      .then((res) => console.log(res.data))
+      .catch((err) => console.log(err));
+    setEventActive(!eventActive);
```
alvyjudy: I wonder if it's better to put this line (`setEventActive()`) into the `.then` block during our development. That way we guarantee that the request is completed before changing the UI. If this route goes wrong, it might be hard to notice on the front end.

It's not really a pressing issue though and we can leave it as it is.

alvyjudy (approved):
Looks really good nice work!

client/src/components/SelectTabs/EventTypesTab.js
```diff
@@ -238,6 +238,7 @@ export default function EventTypesTab() {
         setUserURL(res.data);
       })
       .catch((err) => console.log(err));
+      console.log(userEvents)
```
dproc96: Be sure to remove these console.logs

client/src/components/UserEvents/EventCard.js
dproc96: I agree with this idea, it would be good to know if something went wrong because otherwise you could end up having the exact opposite state as the db

client/src/components/UserEvents/EventGrid.js
```diff
@@ -19,7 +19,7 @@ const useStyles = makeStyles((theme) => ({
 export default function EventGrid({ userEvents }) {
   const classes = useStyles();
   const [userURL, setUserURL] = useState();
-
+  // console.log(userEvents)
```
dproc96: This can be removed

client/src/pages/Scheduler.js
```diff
@@ -64,95 +67,100 @@ const Scheduler = () => {
...
-            />
-          </Route>
-        ) : (
-          <>
-            <Grid item xs={4}>
-              <Overview
-                {...eventDetails}
-                appointmentTime={appointmentDetails.time}
+  if (!eventActive) {
+    return <EventNotActivePage />;
```
dproc96, line 82: I'm seeing a flash of the scheduler page before being redirected, maybe you could use an isLoading state to wait until you know the event is active or not

server/db/models/EventType.js
```diff
@@ -25,6 +25,9 @@ const schema = new Schema({
   members: {
     type: [User],
   },
+  active: {
+    type: Boolean,
```
dproc96: I think it would make more sense for user experience if the default value for this is true

dproc96 (commented):
Overall I love the clean approach to this feature, just a couple of minor notes I added on to Nellie and Di's thoughts

client/src/components/UserEvents/EventCard.js
```diff
@@ -41,24 +59,56 @@ export default function EventCard({ name, duration, color, link, url }) {
...
+                control={
+                  <Switch
+                    size="small"
+                    color="primary"
+                    checked={eventActive}
+                    onChange={handleSwitch}
+                  />
+                }
+                label="On/Off"
+              />
```
sina-jamshidi, line 99: 👍

client/src/pages/EventNotActivePage.js
```diff
@@ -0,0 +1,83 @@
...
+                our website
+              </Link>{" "}
+              to learn more.
+            </Typography>
+          </Grid>
+        </Paper>
+      </Container>
+    </div>
+  );
+}
```
sina-jamshidi, line 83: Nice!

sina-jamshidi (approved):
This looks good to merge once comments are addressed!

client/src/pages/Scheduler.js
WhoaNellie: We also have this problem with the onboarding page flashing before it detects that you've been onboarded, so whatever solution you end up doing for this we should steal for the onboarding page!

server/db/models/EventType.js
mattchx: Sure, I did that originally but removed it.

client/src/components/UserEvents/EventCard.js
mattchx: I was having an async issue with setState so I did it like this but I knew it is not optimal! I can make the changes!

client/src/pages/Scheduler.js
mattchx: Yes, that's valid. But I think for simplicity I will keep it as is unless there is a performance issue.

client/src/pages/Scheduler.js
mattchx: @WhoaNellie I corrected this and I think solution is effective and simple. Check it out.

PR #108, opened 2021-02-01
Team Creation

### LINKED ISSUE #86: BE: Team Management Routes
We'll need a few routes to:

- Create a team
- Update team info
- Add a member to a team
- Remove a member from a team
- Promote a member to an admin (or demote them from admin)

### LINKED ISSUE #90: IN: Create a Team

WhoaNellie:
Basic team routes laid out and team creation integrated. 

Added suggested emails based on all users in db
image

And modified UI to only send invite to members that aren't signed up w/ calendApp (email sending not yet integrated)
image
image


server/routes/teamAPI.js
```diff
@@ -0,0 +1,60 @@
...
+    .then((response) => res.send(response))
+    .catch((error) => {
+      console.log(error);
+      res.status(500).send(error);
+    });
+});
+
+// ADD Team member
+router.get("/api/team/add", auth, (req, res) => {
+  db.Team.updateOne({ _id: req.body.teamId }, { $push: req.body.member })
```
dproc96: Shouldn't this be { $push: { members: req.body.member } })?

server/routes/teamAPI.js
```diff
@@ -0,0 +1,60 @@
...
+        .catch((error) => {
+          console.log(error);
+          res.status(500).send(error);
+        });
+    })
+    .catch((error) => {
+      console.log(error);
+      res.status(500).send(error);
+    });
+});
```
alvyjudy: If time permits, you can make the chain more readable by returning a promise:
```
db.User.updateOne()
.then(()=> db.User.updateOne())
.then(()=>{res.send()})
.catch(()=>{res.send()})
```

alvyjudy (commented):
Nice set up! Is it possible to include some comments on what the data and body would include?
something like `//req.body: {_id, eventId,}` `//data: {newAdminId, members}` etc.

That could save everyone time to add a console log to the route to examine what the data contains.

server/routes/userAPI.js
```diff
@@ -1,18 +1,18 @@
-const express = require('express');
-const db = require('../db/models');
-const auth = require('../middleware/auth');
-const mongoose = require('mongoose');
```
mattchx, line 4: single quotation marks causing havoc

server/routes/teamAPI.js
```diff
@@ -0,0 +1,60 @@
...
+      .catch((error) => {
+        console.log(error);
+        res.status(500).send(error);
+      });
+  });
+});
+
+// Change Admin
+router.post("/api/team/admin", auth, (req, res) => {
+  db.User.updateOne({ _id: req.body.newAdminId }, { $set: { isAdmin: true } })
```
mattchx, line 62: I've read in the documentation that `$set` is implied here however for readability it probably is best to keep it.

server/routes/teamAPI.js
mattchx: @alvyjudy You're recommendation is to add a second `.then` onto the first `.updateOne`. And the second `.then` handles the response of the second `.updateOne`, correct? 
Would that include only one `res.status(500).send(error);`?

server/routes/teamAPI.js
mattchx: I'm thinking the same unless team members are somehow pushed directly onto the array.

server/routes/teamAPI.js
alvyjudy: Yep. Unless we choose to do something differently depending on which error we are dealing with, we can just chain the `then` and have only one `catch((error)=>{res.send(error)}` at the end. It's just to make things more readable though so not a pressing issue lol

server/routes/teamAPI.js
WhoaNellie: Yes, it should!

server/routes/teamAPI.js
```diff
@@ -0,0 +1,73 @@
...
+const auth = require("../middleware/auth");
+
+const router = express.Router();
+
+// CREATE team
+// req.body: { name: name, members: [email:string] }
+router.post("/api/team/create", auth, async (req, res) => {
+  let team;
+  const findMembers = req.body.members.map((email) => {
+    return db.User.findOne({ email: email });
```
kozaktar, line 12: we can use db.User.find({eamil:{$in: req.body.members}}) instead of map

kozaktar (approved):
Good work Nellie!

server/routes/teamAPI.js
WhoaNellie: Wish I knew about this before I made it an array of promises 🙃

PR #109, opened 2021-02-01
Feature/past appointments

alvyjudy:
# Overview

When user navigates to the "scheduled appointment" tab, it fetches from backend an array and separate them into past and upcoming appointments and renders them.

# Added frontend components

- `ScheduledAppointmentsTab.js`: Fetches an array of appointment object from the backend and divide them into upcoming and past by comparing the appointment date and current date.
- `IndividualAppointment.js`: The expandable card-like UI component that displays each appointment
- `YourAppointments.js`: Text divider indicating whether the appointments are upcoming or past, empty or not.

# Added and modified backend routes 

`routes/appointmentAPI.js`, `db/models/appointment.js`

- POST `/api/appointment`: I removed the `auth` middleware because the **attendee is not gonna be authenticated when they schedule an appointment**. This route first queries the `EventTypes` model,  gets the id of the user that's hosting the event (`userId`), and includes it as `hostId` in the `appointment` model
- GET `/api/all-appointments`: queries the `hostId` key in `appointment` model via userId (in `req.userId`) and returns an array of appointments
- Added a `hostId` field in the `appointment` model

# Limitations

- Does not display appointment duration, colour flag and other useful information. This is because the `appointment` model does not contain those information (I only added the necessary `hostId` key). I'm debating whether I should: 1) have each component (`IndividualAppointment.js`) grab those information from the `EventType` model upon render, or 2) include such information in the `appointment` model upon appointment creation. Any suggestions?
- I have yet to implement the reschedule and cancel button action. Would like some more inputs on what should be included in the UX/UI so I can organize things better.

# UI screenshots
image
image


client/src/components/ScheduledAppointments/ScheduledAppointmentsTab.js
```diff
@@ -0,0 +1,96 @@
...
+
+import { useUserData } from "../../providers/Context";
+import IndividualAppointment from "./IndividualAppointment";
+import YourAppointments from "./YourAppointments";
+
+const ScheduledAppointmentsTab = () => {
+  const userData = useUserData();
+  const { _id: userId } = userData;
+
+  const [upcomingAppointments, setupcomingAppointments] = useState([]);
```
dproc96:
```suggestion
  const [upcomingAppointments, setUpcomingAppointments] = useState([]);
```
You'll have to change this in other places too

client/src/components/ScheduledAppointments/ScheduledAppointmentsTab.js
```diff
@@ -0,0 +1,96 @@
...
+            attendeeTimezone={appointment.timezone}
+          />
+        ))}
+    </Grid>
+  );
+};
+
+const appointmentIsUpcoming = (appointmentDateInISOString) => {
+  // upcoming === true indicate that the appointment is either today or tomorrow onward
+  const today = startOfToday();
```
dproc96: Should we compare to the start of today or to the current time?

alvyjudy: I might have compared it to the start of today to keep things simple. But yeah it makes more sense to compare to the current time.

client/src/components/ScheduledAppointments/ScheduledAppointmentsTab.js
```diff
@@ -0,0 +1,96 @@
...
+          upcomingAppointments.push(appointment);
+        } else {
+          pastAppointments.push(appointment);
+        }
+      });
+
+      setupcomingAppointments(upcomingAppointments);
+      setPastAppointments(pastAppointments);
+    });
+  }, [userId]);
```
mattchx: Why is `userId` the dependency here?

client/src/components/ScheduledAppointments/ScheduledAppointmentsTab.js
mattchx: It depends... on what we consider upcoming(tomorrow vs in a couple hours) and also how `compareAsc` calculates the comparison.

server/db/models/Appointment.js
```diff
@@ -6,6 +6,9 @@ const schema = new Schema({
   eventId: {
     type: mongoose.Schema.Types.ObjectId,
   },
+  hostId: {
```
mattchx: `hostId` would be the same number id as `eventId`?

server/routes/appointmentAPI.js
```diff
@@ -1,27 +1,37 @@
...
+  try {
+    const { eventId } = req.body;
 
-  db.Appointment.create({ ...req.body })
-    .then((response) => res.send(response))
-    .catch((error) => {
-      console.log(error);
-      res.status(500).send(error);
+    //get the userId that's hosting this event
+    const { userId: hostId } = await db.EventType.findOne({ _id: eventId });
```
mattchx: You are getting the `hostId` and setting that to an object with a key of `userId ` correct?

mattchx (commented):
Great work once again!

client/src/components/ScheduledAppointments/ScheduledAppointmentsTab.js
alvyjudy: Yeah Matt we definitely should discuss that. I have set up the function so I can easily adjust what we consider as upcoming. The component only need to know whether it's true or false lol.

client/src/components/ScheduledAppointments/ScheduledAppointmentsTab.js
alvyjudy: Actually I can probably omit that (which saves a few lines of code since I'm not using other parts of user data lol). The reason though is because each request is associated with the specific userId (via cookies token) so it should be a dependency. I think it's more of a theoretical best practice lol.

server/db/models/Appointment.js
alvyjudy: Oh no `hostId` is the `userId`. I wanted to include it as `userId` but thought that would be confusing because for each appointment, there is a host and attendee, and it's not so intuitive to decide which one is the "user". 

But I don't know if I ended up making it more confusing lol. Let me know how you feel and we can pick the one that make more sense to everyone.

server/routes/appointmentAPI.js
alvyjudy: Just renaming. `const {userId: hostId} = await ...` is equivalent to:
```
const {userId} = await ...;
const hostId = userId;

File: server/db/models/Appointment.js (hunk 4 above)
Comment (mattchx) [reply server/db/models/Appointment.js]: To me, it works like this but we can discuss if there are any suggestions. 
A simple explation would be: the hostId is set to userId of the current host.

Comment (alvyjudy) [reply server/db/models/Appointment.js]: That's actually a great explanation lol I should have included that in the comment

**Comment by sina-jamshidi** [comment]
> # Overview
> When user navigates to the "scheduled appointment" tab, it fetches from backend an array and separate them into past and upcoming appointments and renders them.
> 
> # Added frontend components
> * `ScheduledAppointmentsTab.js`: Fetches an array of appointment object from the backend and divide them into upcoming and past by comparing the appointment date and current date.
> * `IndividualAppointment.js`: The expandable card-like UI component that displays each appointment
> * `YourAppointments.js`: Text divider indicating whether the appointments are upcoming or past, empty or not.
> 
> # Added and modified backend routes
> `routes/appointmentAPI.js`, `db/models/appointment.js`
> 
> * POST `/api/appointment`: I removed the `auth` middleware because the **attendee is not gonna be authenticated when they schedule an appointment**. This route first queries the `EventTypes` model,  gets the id of the user that's hosting the event (`userId`), and includes it as `hostId` in the `appointment` model
> * GET `/api/all-appointments`: queries the `hostId` key in `appointment` model via userId (in `req.userId`) and returns an array of appointments
> * Added a `hostId` field in the `appointment` model
> 
> # Limitations
> * Does not display appointment duration, colour flag and other useful information. This is because the `appointment` model does not contain those information (I only added the necessary `hostId` key). I'm debating whether I should: 1) have each component (`IndividualAppointment.js`) grab those information from the `EventType` model upon render, or 2) include such information in the `appointment` model upon appointment creation. Any suggestions?
> * I have yet to implement the reschedule and cancel button action. Would like some more inputs on what should be included in the UX/UI so I can organize things better.
> 
> # UI screenshots
> image
> 
> image

It would be nice to get the rest of the details as well from the EventType model. Also UI/UX notes:
- can we still show the headings ("Upcoming Appointments", "Past Appointments") even if none are scheduled? Just write the "no upcoming appointments scheduled" under the heading.
- The rest of team can weigh in on this, but technically the "reschedule" and "cancel" options are for the attendee, not the organizer, so maybe we can make it all cleaner by removing the need for the accordion and just having simple cards with the time, attendee email, timezone, duration, event type name, and event type colour in a thin bar

File: client/src/components/ScheduledAppointments/YourAppointments.js [hunk 6]
```diff
@@ -0,0 +1,27 @@
...
+      </Typography>
+    );
+  } else {
+    return (
+      <Typography className={classes.text} variant="h4">
+        Your {upcoming ? "upcoming" : "past"} appointments
+      </Typography>
+    );
+  }
+};
```
sina-jamshidi: As mentioned in comment on PR, maybe leave the heading always there, and just make the "you do not have any..." conditional

server/db/models/Appointment.js
sina-jamshidi: I think an even more clear solution is just to call it `hostUserId` lol

alvyjudy:
# Most recent changes
- appointment component uses card instead of accordion
- always display the header for "Your upcoming appointment", "You do not have any appointment", etc.
- Added a backend route to fetch event details via eventId (I find this way cleaner because the eventId is already included in the appointment database model)
- Each appointment card fetches from the `/api/event-details-via-id/:eventId` for information such as duration, color, etc.
- appointment model uses key `hostUserId` instead of `userId`.
- The appointment time is greyed out when it's in the past
- Accurately determine whether the appointment is in the past or not by comparing its time to the current time (with `Date.now()`). If the appointment has not occurred, it is considered upcoming.

Let me know if there are more suggestions!

# Screenshots
Screen Shot 2021-02-01 at 6 38 33 PM

image
[](url)

client/src/components/ScheduledAppointments/IndividualAppointment.js
```diff
@@ -0,0 +1,98 @@
...
+  eventId,
+}) => {
+  const parsedDateObj = parseISO(time);
+  const formattedTime = format(parsedDateObj, "h:mm b");
+  const formattedDate = format(parsedDateObj, "E, LLL do yyyy");
+
+  const [eventDetails, setEventDetails] = useState({});
+  const classes = useStyles({ eventColor: eventDetails.eventColor, upcoming });
+  useEffect(() => {
+    axios.get(`/api/event-details-via-id/${eventId}`).then(({ data }) => {
```
dproc96: Can you think of a more efficient way to do this?

server/db/models/Appointment.js
```diff
@@ -6,6 +6,9 @@ const schema = new Schema({
   eventId: {
     type: mongoose.Schema.Types.ObjectId,
   },
+  hostUserId: {
+    type: String, // Using string type here to prevent type conversion error since all we care if whether a given id value matches it
```
dproc96: I see your point in the comment, but wouldn't it make sense to make this a reference anyway for potential future applications? You can just use `mongoose.Types.ObjectId(str)` to deal with conversion issues (not positive on that exact syntax)

client/src/components/ScheduledAppointments/IndividualAppointment.js
alvyjudy: uhh I'm not sure I understand which direction I should aim for. Did you mean a more efficient way to fetch the data or handle the response?

dproc96: Fetching the data

server/db/models/Appointment.js
alvyjudy: I used string because I still want the server to respond with a "id not found" instead of "casting error", when frontend sends a invalid string like "abcd1234". The conversion is done automatically as long as the string is of 24 hex characters. 

Should I just have the route check for the string format before actually letting it hit the database?

sina-jamshidi (approved):

PR #123, opened 2021-02-02
Add loading state to context

alvyjudy:
# Summary 

This PR should pave way for our bug fixes involving screen flash and loading spinner.

- Consolidate the auth context. Instead of fetching the authentication state in `App.js`, it now does it in the context provider.
- Add a `dataLoading` and `authLoading` state into context, and their corresponding hooks.

## Hooks

```
import {useAuthLoading, useDataLoading} from "./providers/Context";


const Component = () => {
  const authLoading = useAuthLoading(); // true or false
  const dataLoading = useDataLoading(); // true or false

  // render spinner if authLoading or dataLoading is true. Render actual content when false.
}

### DIFF
--- a/client/src/App.js
+++ b/client/src/App.js
@@ -13,24 +13,13 @@ import UpgradePage from "./pages/UpgradePage";
 import PageNotFound from "./pages/PageNotFound";
 import PrivateRoute from "./components/Routes/PrivateRoute";
 import Checkout from "./pages/Checkout";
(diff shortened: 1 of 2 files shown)


dproc96 (approved):
This looks good and should make an easy fix for a lot of situations! There may be future examples of this issues that aren't resolved by this fix (for example, the one Matt fixed in his PR) but I think in those cases you can just use a basic loading state

PR #125, opened 2021-02-02
Bugfix/redirect to home page

### LINKED ISSUE #110: Handle Expired Tokens
Right now if you load the home page with an expired token you just see an empty version of the homepage, it should redirect to login

alvyjudy:
Fixes #110. This branch is based on #123 . The file modified is `src/components/Routes/PrivateRoute.js`

`<PrivateRoute/>` now checks for both loading state and authentication state via the two hooks.

- `loading === true`: do nothing
- `loading === false` & `authenticated === false`: redirect to "/login"
- `loading === false` & `authenticated === true`: render the `children` props

### DIFF
--- a/client/src/components/Routes/PrivateRoute.js
+++ b/client/src/components/Routes/PrivateRoute.js
@@ -1,11 +1,19 @@
 import React from "react";
 import PropTypes from "prop-types";
-import { Route } from "react-router-dom";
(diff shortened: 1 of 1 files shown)


dproc96 (approved):
"/" should also be a PrivateRoute, otherwise this looks great!

PR #133, opened 2021-02-02
Check user email on login

kozaktar:
Features:
-Added a backend route to check if email is already registered in the app
-Errors are now displayed on login/signup if the user does (signup) or doesn't exist in the app db (login)

https://user-images.githubusercontent.com/20387499/106651400-18128780-6562-11eb-86ee-5c44f37ed467.mp4


dproc96 (commented):
Be sure you are running prettier on your files to keep your formatting consistent. Also hold off on merging this until Matt's changes are merged in, otherwise this is good

PR #136, opened 2021-02-02
Improve card styling and add delete eventType functionality

mattchx:
### COMPLETE** with bug fix!

https://user-images.githubusercontent.com/55170649/106833101-f3e9a000-6660-11eb-900b-cc4be68175f3.mov

### UPDATE

https://user-images.githubusercontent.com/55170649/106832735-3a8aca80-6660-11eb-8601-44441e96b35b.mov

### First Draft
- cards now have larger box shadow on hover
- non active event cards show a new 'turn on' button and switch is hidden in this state
https://user-images.githubusercontent.com/55170649/106658548-71cb7f80-656b-11eb-9418-970fbe5b6f41.mov


WhoaNellie (commented):
I think it might make more sense to leave the toggle for switching on/off but remove the on/off text and replace it with a tooltip that says "disable event" or "enable event" depending of what state the toggle is it 
https://material-ui.com/components/tooltips/

mattchx:
@WhoaNellie I'm willing to incorporate the tooltip and maybe keep the button. I've basically took ideas from calendly. I have added a timeout to show the switch animation before the switch disappears . We can have a vote on this. I'm flexible.

WhoaNellie:
@mattchx I think the reason the toggle/button combo is strange to me is that using one position on a toggle functionally makes it a checkbox/radio button that has to be unchecked by the "Turn on button", so I wonder why it's a toggle instead of something else

dproc96 (approved):
I would agree with Nellie about the turn on button being a little strange for user experience. If I use a switch to turn something off I expect to use the same switch to turn it back on.

I love the hover effect though, it adds a lot of slickness to the app!

client/src/components/SelectTabs/EventTypesTab.js
```diff
@@ -240,6 +240,13 @@ export default function EventTypesTab() {
       .catch((err) => console.log(err));
   }, []);
 
+  const updateUserEvents = (eventId) => {
+    console.log('update events')
```
kozaktar: remove console.log

client/src/components/SelectTabs/EventTypesTab.js
```diff
@@ -240,6 +240,13 @@ export default function EventTypesTab() {
...
+    setUserEvents(userEvents.filter((event) =>
+      event._id !== eventId && event
```
kozaktar: can probably replace by just  event._id !== eventId . Since event._id will throw an error anyways if event is null.

kozaktar: or if you are trying to see if event is null then it should be reversed:
event && event._id !== eventId
that way if event is null then second part won't be checked.

client/src/components/UserEvents/EventGrid.js
```diff
@@ -27,12 +27,22 @@ export default function EventGrid({ userEvents }) {
     });
   }, []);
 
+  const handleDelete = (eventId) => {
+    axios
+      .delete("/api/team-event/delete/", {data: { id: eventId}})
+      .then((data) => {
+        console.log(data)
```
kozaktar: remove console.log

kozaktar (approved):
Looks good Matt. Just remove console.logs and its good to go!

client/src/components/SelectTabs/EventTypesTab.js
mattchx: Let's discuss this on slack

dproc96:
Screen Shot 2021-02-04 at 9 12 03 AM

The popup displays strangely on my screen, can you fix this before merging?

client/src/components/SelectTabs/EventTypesTab.js
mattchx: `event._id !== eventId ` this is the correct simplification since it's the event is returned when this condition it true

mattchx:
Made all requested changes!

PR #137, opened 2021-02-02
Optimize data fetching in feature/past-appointments

alvyjudy:
This PR addresses Dan's [comment](https://github.com/hatchways/team-yugioh/pull/109#discussion_r568620184) and today's discussion on making the data flow more efficient (I'm opening a new one here so it's easier to review)

- When querying `/api/all-appointments`, the backend do all the hard lifting to combine the details for the appointment and event type and send it to front end.
- The eventdetails is obtained by calling the function `getEventDetailsViaId(eventId)`, which can later be hooked up to a cache object (I can implement that if this set up looks good to everyone).
- `<IndividualAppointment/>` no longer needs to do its own data fetching

### DIFF
--- a/server/routes/appointmentAPI.js
+++ b/server/routes/appointmentAPI.js
@@ -42,6 +42,7 @@ router.post("/api/appointment", async (req, res) => {
 //   link, // the incomplete link for building invitation link: <hostName>/<eventName>
 //   members,
 //   active, // indicate whether the event is active or not
+//   eventName
 // }, ...]
 router.get("/api/all-appointments", auth, async (req, res) => {
   try {
--- a/server/routes/appointmentAPI.js
+++ b/server/routes/appointmentAPI.js
@@ -2,6 +2,7 @@ const express = require("express");
 const db = require("../db/models");
 const auth = require("../middleware/auth");
 const ensureAppointmentExists = require("../middleware/ensureAppointmentExists");
+const getEventDetailViaId = require("../utils/getEventDetailsViaId");
 const router = express.Router();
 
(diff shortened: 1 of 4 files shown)


server/routes/appointmentAPI.js
```diff
@@ -25,17 +26,61 @@ router.post("/api/appointment", async (req, res) => {
...
+    const appointmentsWithEventDetailsPromise = appointments.map(
+      async (appointment) => {
+        const eventDetails = await getEventDetailViaId(appointment.eventId);
+        // Order matters, both of them contain a `name` key,
+        // in eventDetails, that's the event's name,
+        // in appointment, that's the attendee's name
+        // I will save only the attendee's name
+        // Also, mongoose query returned object is not a regular object
+        // need to convert before spreading
+        // big pain
```
dproc96: FYI large comments like this explaining approach are often frowned upon. Not really worth changing but for the future when you feel a comment is necessary try to either get the point across with variable names or if that's not possible then try to keep the comment to one line like `// spread eventDetails first because of duplicate keys`

server/utils/getEventDetailsViaId.js
```diff
@@ -0,0 +1,9 @@
+const db = require("../db/models");
+
+// Returns a promise that resolves to an object containing the details of the
+// event. This function can be optimized by hooking it onto a cache object
```
dproc96, line 4: Good eye to the future, I think having a cache for event details makes a lot of sense

dproc96 (approved):
This looks good to me, nice work setting up the potential for caching

PR #143, opened 2021-02-03
users can have spaces in urls now

### LINKED ISSUE #121: Bug: handle white space in URL
It looks like the URL is still valid when it contains white space, which leads to issue when browser sees it. Shouldn't be a hard fix but there seem to be multiple approach to it.

image
image

WhoaNellie:
Encoding the URI in a few places so users will be able to put spaces in their links

### DIFF
--- a/client/src/components/SelectTabs/EventTypesTab.js
+++ b/client/src/components/SelectTabs/EventTypesTab.js
@@ -296,7 +296,7 @@ export default function EventTypesTab() {
       .post("/api/event", {
         ...eventBody,
         duration: minutes,
-        link: `${userURL}/${eventBody.link}`,
+        link: encodeURI(`${userURL}/${eventBody.link}`),
       })
       .then((res) => {
         const currentEventTypes = [...userEvents];
@@ -342,7 +342,6 @@ export default function EventTypesTab() {
             </Button>
           </Box>
         </Grid>
-        {/* TYPE OF EVENT CARDS */}
         <EventGrid userEvents={userEvents} />
       </Container>
       <Dialog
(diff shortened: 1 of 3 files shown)


client/src/pages/Scheduler.js
```diff
@@ -54,9 +54,11 @@ const Scheduler = () => {
 
   useEffect(() => {
     // Matching only the first two params so /hostname/eventname/datelinks will still work correctly
-    const regex = /appt\/([0-9a-z-]*\/[0-9a-z-]*)/g;
+    const regex = /appt\/([%\s0-9a-z-]*\/[%\s0-9a-z-]*)/g;
```
alvyjudy, line 57: I wonder if we should just grab the string and run a `decodeURIComponent()` on it. But if this works then let's just go with it.

client/src/components/SelectTabs/EventTypesTab.js
```diff
@@ -296,7 +296,7 @@ export default function EventTypesTab() {
       .post("/api/event", {
         ...eventBody,
         duration: minutes,
-        link: `${userURL}/${eventBody.link}`,
+        link: encodeURI(`${userURL}/${eventBody.link}`),
```
alvyjudy, line 299: The proper function should be encodeURIComponent I think. But this should work as well.

alvyjudy (approved):
I left some suggestions but they aren't critical. If it works let's go with what you have already!

client/src/pages/Scheduler.js
WhoaNellie: This works but that could possibly be a better way!

client/src/components/SelectTabs/EventTypesTab.js
WhoaNellie: Link isn't a complete uri, it's just the user's url pref and suffix so this should be fine I believe!

PR #151, opened 2021-02-03
Add object id format check in feature/past-appointments

alvyjudy:
- `hostUserId` uses `ObjectId` type in database
- Added format check for the object Id and send a "invalid format" error 
- It appears that the validate function from mongoose library function only checks whether the string is 12 characters long (future improvement)


dproc96 (approved):
Adding a function that throws an error and then catching and handling the error feels like a complex approach versus just checking the validity and sending an error if it's false, but it's not too much extra code so I think this is fine

PR #155, opened 2021-02-03
Integrate set availability in onboarding 

alvyjudy:
- This PR let the onboarding component post the available time and day to the database
- Days are now stored in an array whose number denotes day of the week. `0` is Sunday, `6` is Saturday.
- Solves the warning that says "children props received null" in `SetTimezoneUrl`

I will open another PR to actually use these availability data as that require more work than I anticipated.

### DIFF
--- a/server/db/models/User.js
+++ b/server/db/models/User.js
@@ -29,6 +29,12 @@ const schema = new Schema({
   URL: {
     type: String,
   },
+  availableTime: {
+    type: Mixed, // {start: "HH:MM", end: "HH:MM"}
+  },
+  availableDays: {
+    type: [Number], // Number indicates day of week. 0: Sunday, 6: Saturday
+  },
   subscribed: {
     type: Boolean,
   },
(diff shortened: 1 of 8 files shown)


kozaktar (changes requested):
Can you also remove the edit buttons in onboarding dialog since they don't do anything?

alvyjudy:
Good point Taras!

kozaktar (commented):
Looks good Di!

kozaktar (approved):

server/db/models/User.js
```diff
@@ -29,6 +29,12 @@ const schema = new Schema({
   URL: {
     type: String,
   },
+  availableTime: {
+    type: mongoose.Schema.Types.Mixed, // {start: "HH:MM", end: "HH:MM"}
```
dproc96, line 33: You could also define this as such:
```javascript
availableTime: {
  start: String,
  end: String
}
```

dproc96 (approved):
Other than my very small suggestion, this looks good to go!

alvyjudy:
@WhoaNellie I have made the edits to store timezone as integer! Please take a look at line 26 on `OnBoarding.js` and line 147 on `SetTimezoneUrl.js`

PR #156, opened 2021-02-04
fixed check link in onboarding and new events dialog

kozaktar:
It seems at some point debounceCheckUnique was changed from 
debounceCheckUnique = useCallback(debounce(checkUnique, 500), []);
to:
 const debounceCheckUnique = useCallback(() => {
    debounce(checkUnique, 500);
  }, []);
  
  which seems to have broken in.


dproc96 (approved):
This looks good, nice fix!

PR #159, opened 2021-02-04
Feature/embed scheduler in another website

alvyjudy:
Thanks to Sina's idea, I implemented another cool feature that allows other developers to embed our scheduler into their website!

# Overview

1. On the event card, there is a button that shows a block of code, which the user can copy and paste into their website:

Screen Shot 2021-02-04 at 3 05 41 AM

*** 

image

2. User then copy the code onto their other website (here I assume just html) and they will be able to see the scheduler being displayed

image

***

image

# How it works

The `<script>` links to a route on the backend, which sends back a javascript file containing the code to render the scheduler. This code is generated by a separate webpack build set up (inside `client/widget-build`) and stored in `server/routes/assets/widget`. 

# Instruction

- Navigate to `client/widget-build` and run `yarn run webpack --config widget.config.js`. This will generate the bundled scheduler that will be referenced with `<script src="...">`

# Limitation & room for improvement
- build pipeline need to be integrated with the main app
- domain name needs to be set in a more sensible way. Right now I just hardcoded `http://localhost:3000` to it.
- styles can be more refined
- there is a fair (read: huge) amount of repeated code between `Scheduler.js` and `EmbeddedScheduler.js`
- availability still doesn't work because the user isn't authenticated


client/src/widget/EmbeddedScheduler.js
```diff
@@ -0,0 +1,184 @@
+// Esssentially a copy of the `Scheduler.js` component
```
dproc96, line 1: Is there a way to just use the Scheduler component (maybe with some extra props) to avoid repeating code?

client/src/widget/EmbeddedScheduler.js
```diff
@@ -0,0 +1,184 @@
...
+const domain = "http://localhost:3000";
```
dproc96: You should move this down to line 20 to avoid `import in body of module` errors, also you may be able to set this to `window.location.origin` to make your life easier when deploying (that also may not work so let me know if I'm wrong here)

client/src/widget/EmbeddedScheduler.js
```diff
@@ -0,0 +1,184 @@
...
+                direction="column"
+                wrap="nowrap"
+                spacing={2}
+              >
+                {appointmentDetails.time ? (
+                  <AppointmentDetails
+                    appointmentDetails={appointmentDetails}
+                    setAppointmentDetails={setAppointmentDetails}
+                    eventDetails={eventDetails}
+                    path={path}
```
dproc96: I'm getting an error that path is not defined

dproc96 (commented):
This looks absolutely incredible, I had a couple of warnings that I saw and one idea to address the domain issue you noted in your comment.

sina-jamshidi (approved):
Looks good to go! It would just be nice if you could make the code snippet be in a box with an outline, and the copy button styled nicer to be next to it.

alvyjudy:
Here is what it looks like now: 
image

And what it looks like when applied to some random html template:
image

client/src/widget/EmbeddedScheduler.js
alvyjudy: Turns out this value needs to be set at build time to send request to the proper location. I think I will just stick to localhost for now lol

PR #167, opened 2021-02-04
Feature/responsive nav bar

mattchx:
https://user-images.githubusercontent.com/55170649/106956749-0c12fb00-6705-11eb-85d4-2f59a2730a50.mov


alvyjudy:
Looks great to me! Could you sync it with dev and then go through the debugging checklist we have? Then I think this can be merged. Let me know when you sync it I will run through the checklist too

PR #173, opened 2021-02-05
Bugfix/onboarding misc

### LINKED ISSUE #165: New users' links aren't being populated on the page
When you sign up a new user and they first go to the dashboard, their link on the EventTab and New Event Dialog show as blank

### LINKED ISSUE #170: Can't uncheck days on onboarding avilavbility
image

WhoaNellie:
Fixed state causing issues with the checkboxes on onboarding and made it so that the user context is updated after they're onboarded.

Probably a better way to handle the checkbox state but this seems to work for now.

### DIFF
--- a/client/src/components/onboarding/AvailableDays.js
+++ b/client/src/components/onboarding/AvailableDays.js
@@ -1,4 +1,4 @@
-import React from "react";
+import React, { useState } from "react";
 import {
   Checkbox,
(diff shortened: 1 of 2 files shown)


alvyjudy (approved):
That's actually a cool way to do the check days part! I tested it briefly and I think this is good to merge!

PR #176, opened 2021-02-05
Bugfix/add customer name and email to stripe record

alvyjudy:
Currently there's something on the payment API so the user email and name isn't stored on Stripe. Fixed it with this PR


WhoaNellie (approved):
Looks good!

PR #183, opened 2021-02-15
Deployment!

alvyjudy:
App is live at: https://calendy-team-yugioh.herokuapp.com/ 
Currently I manually deployed against the `deployment` branch at commit hash "a1b0b6". We can configure it to automatically deploy against the default branch after this PR is approved by everyone.

# Summary of changes

### Set up yarn workspace to manage build, start steps and dependency management

The heroku workflow would run 1) `yarn install`,  2) `build` and 3) `start` script based on the `package.json` in the root folder. Using yarn workspace allows us to orchestrate the two subfolders in a concise manner. 

I wrote a [post](https://alvyjudy.github.io/2021/02/14/heroku-mono-repo.html) to document the process if anyone's interested 😜

### Use `window.location.origin` to construct urls instead of hardcoding

The embedding part still needs some work. I haven't got to it yet.

### Added a demo button to get access to the app 

Internally, the button sends request to the newly added "sign up" route (see below) with a randomly generated email address and password. The button is a bit ugly, need a better idea as to where to put it lol.

### Added backend route for sign up and login with password

Serve as a fallback when something goes wrong with google auth and for easier demo 

### Availability would bypass google calendar if it's not available 

Right now it just checks the presence of token, could be set up better. If the token is absent, use the availability from the onboarding.


kozaktar (commented):
Looks good Di. Maybe we can put the Demo button under login button and also add tooltip "no google calendar integration" to it to inform the user that that feature its not available in demo. Or we can have a google account that the app uses for demo.

alvyjudy:
Thanks Taras! I put the button underneath login. Didn't do the tool tip though. If you want to open a PR for that feel free to do so!

